PR #3, opened 2024-02-14
Dump thanos metrics to csv

AlbertoSoutullo:
Small PR to slowly give you a vision on how the repo is going.

I am starting with the metrics scrapping since the deployment is working, and the metrics are the only “problem” right now.
Each metric should have its own `template`, but I am still not sure if we will inject this in a Helm chart, or we add is as a some kind of YAML config.

What this does is query Thanos, prepare the data as a .csv, and dump it locally. I will add tests when we have a better direction of how to settle the configuration, since this can still vary a lot.

@AlejandroCabeza please, feel free to be as picky as you want.


src/metrics/scrapper.py
```diff
@@ -1,32 +1,125 @@
...
+            output_dir.mkdir(parents=True, exist_ok=True)
+        except OSError as e:
+            return Err(f'Error creating {output_dir}. {e}')
+
+        return Ok(output_dir / output_file)
+
+    def _create_dataframe_from_data(self, data: Dict) -> pd.DataFrame:
+        final_df = pd.DataFrame()
+        for pod_result_dict in data['result']:
+            column_name = pod_result_dict['metric']['pod'] + '_' + pod_result_dict['metric']['node']
```
AlejandroCabeza, line 85: I'd suggest using f-strings as much as possible as the string formatting standard.
```suggestion
column_name = f"{pod_result_dict['metric']['pod']}_{pod_result_dict['metric']['node']}"
```

src/metrics/scrapper.py
```diff
@@ -1,32 +1,125 @@
...
+        return df
+
+    def _create_pod_df(self, column_name, values) -> pd.DataFrame:
+        pod_df = pd.DataFrame(values, columns=['Unix Timestamp', column_name])
+        pod_df['Unix Timestamp'] = pd.to_datetime(pod_df['Unix Timestamp'], unit='s')
+        pod_df.set_index('Unix Timestamp', inplace=True)
+
+        return pod_df
+
+    # TODO this depends on pods name assigned in deployment
```
AlejandroCabeza, line 107: The way you coded it seems safe enough for "initial order" not to matter.

src/metrics/scrapper.py
```diff
@@ -1,32 +1,125 @@
...
-            response = requests.get(promql)
-            logger.info(f"Response: {response.status_code}")
+            logger.info(f'Querying {metric}')
+            promql = self._create_query(metric)
 
+            result = self._make_query(promql)
+            if result.is_err():
+                logger.warning(f'Error querying {metric}. {result.err_value}')
 
+            response = result.ok_value
```
AlejandroCabeza, line 35: You're not exiting the function after the `is_err` check, which means that if there is an error you'll trigger an exception (if that's how the underlying library handles it) when accessing `result.ok_value`.

src/metrics/scrapper.py
AlejandroCabeza, line 1: I see there's some functionality you implemented over pandas' dataframe which is defined as `Scrapper`'s methods. If you wanna be a tad extra, I'd consider subclassing the pandas dataframe and grouping all the related functions in there.

src/metrics/scrapper.py
```diff
@@ -1,32 +1,125 @@
...
+        logger.info(f'{metric} data dumped')
+
+    def _prepare_path(self, metric: str) -> Result[Path, str]:
+        output_file = f'{metric}.csv'
+        output_dir = Path(self._out_folder)
+
+        try:
+            output_dir.mkdir(parents=True, exist_ok=True)
+        except OSError as e:
+            return Err(f'Error creating {output_dir}. {e}')
```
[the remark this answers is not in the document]
AlbertoSoutullo: If I am not wrong, you can still get an `OSError` because of things like broken symlinks or the path already exists but its not a directory. More info [here](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir).

src/metrics/scrapper.py
AlbertoSoutullo: Oh yes, I tend to use that, but I missed this one. Nice catch.

src/metrics/scrapper.py
AlbertoSoutullo: It has nothing to do with initial order. Is actually how you name the deployment in Kubernetes. I.e.: maybe there are no `bootstrap` nodes, or nodes can be called `X`. We are sticking to this right now, but will be configurable later.

src/metrics/scrapper.py
AlbertoSoutullo: Yup, the idea was to add a `continue` to go to the next metric. Thanks.

src/metrics/scrapper.py
AlbertoSoutullo: Yes, this is something I had in mind. “Similar” to the `scrape_utils.py`.

AlbertoSoutullo:
@AlejandroCabeza This is ready for review. 

- Scrapping configuration is now in a file `scrape.yaml`
- Minor refactors
- Added tests

src/metrics/scrapper.py
```diff
@@ -1,32 +1,121 @@
...
+                    logger.info(f'Successfully extracted {metric} data from response')
+                case Err(err):
+                    logger.info(err)
+                    continue
 
+            logger.info(f'Dumping {metric} data to .csv')
+            self._dump_data(metric, column_name, data)
 
+    def _set_query_config(self):
+        self._query_config = read_yaml_file(self._query_config_file)
```
AlejandroCabeza, line 42: It's a good practice to set attributes in the `__init__`, even if it's to `None`. That way you can, at a glimpse, understand all the parts of a class.

src/metrics/scrapper.py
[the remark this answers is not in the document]
AlejandroCabeza: True, didn't consider that.

src/metrics/scrapper.py
AlejandroCabeza: Right, got it. The `order` issue is because I'm still with the `groupby` solution we discussed.

src/utils/file_utils.py
```diff
@@ -0,0 +1,14 @@
...
+# Project Imports
+
+
+def read_yaml_file(file_path: str):
+    path = Path(file_path)
+
+    with open(path, 'r') as file:
+        data = yaml.safe_load(file)
+
+    return data
```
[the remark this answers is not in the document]
AlbertoSoutullo: Afaik, `Path` takes care of handling path separators and other OS-specific details.

src/metrics/scrapper.py
AlbertoSoutullo: True, added.

Also, not a specific one right now. Maybe later we would change this to use several files, but right now there is no specific reason.

src/utils/file_utils.py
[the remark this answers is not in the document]
AlejandroCabeza: Sorry, could've phrased that better. What I wanted to say is, why not directly ask for a `Path` object as the parameter?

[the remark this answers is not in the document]
AlbertoSoutullo: Because this will probably be called either by an argument given by a config file or CLI, so we can use this function directly with that. Don't really see a benefit of extracting that line to outside.

AlejandroCabeza (approved):
LGTM

PR #15, opened 2024-03-20
Add Thanos portforward

### LINKED ISSUE #9: Thanos porftforward
Add Thanos port forward to scrape metrics instead of relying on a hard-coded load balancer. More info:

https://github.com/kubernetes-client/python/issues/2195

AlbertoSoutullo:
Closes #9

### DIFF
--- a/src/metrics/kubernetes.py
+++ b/src/metrics/kubernetes.py
@@ -69,14 +69,13 @@ class KubernetesManager:
         return pods, name
 
     def _find_service_port_name_in_pods(self, port: str, pods: V1PodList) -> int:
-        if isinstance(port, str):
-            for container in pods.items[0].spec.containers:
-                for container_port in container.ports:
-                    if container_port.name == port:
-                        return container_port.container_port
-            else:
-                raise RuntimeError(
-                    f"Unable to find service port name: {port}")
+        for container in pods.items[0].spec.containers:
+            for container_port in container.ports:
+                if container_port.name == port:
+                    return container_port.container_port
+        else:
+            raise RuntimeError(
+                f"Unable to find service port name: {port}")
 
     def _find_pod_in_service(self, dns_name, name, namespace, port) -> Tuple[str, int]:
         if dns_name[1] in ('svc', 'service'):
--- a/src/metrics/kubernetes.py
+++ b/src/metrics/kubernetes.py
@@ -1,0 +1,95 @@
+# Python Imports
+import socket
+import logging
+from kubernetes.client import CoreV1Api
+from kubernetes.stream import portforward
+
+
+logger = logging.getLogger(__name__)
+
+
+class KubernetesManager:
+    def __init__(self, api: CoreV1Api):
+        self._api = api
+
+    def create_connection(self, address, *args, **kwargs):
+        dns_name = self._get_dns_name(address)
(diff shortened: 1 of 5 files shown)


src/metrics/scrapper.py
```diff
@@ -1,36 +1,41 @@
...
         self._url = url
         self._query_config = None
         self._query_config_file = query_config_file
         self._out_folder = out_folder
         self._set_query_config()
-        # TODO make interval match value in cluster
+        self._k8s = kubernetes.KubernetesManager(api)
 
     def query_and_dump_metrics(self):
+        socket.create_connection = self._k8s.create_connection
```
AlejandroCabeza, line 29: I'd suggest leaving a comment so future people (including us) understand why in heavens this is necessary 😆

src/metrics/kubernetes.py
```diff
@@ -0,0 +1,97 @@
...
+import logging
+from typing import List, Tuple
+from kubernetes.client import CoreV1Api, V1PodList
+from kubernetes.stream import portforward
+
+
+logger = logging.getLogger(__name__)
+
+
+class KubernetesManager:
```
AlejandroCabeza, line 12: Given this is sourced from some kubernetes code (as you told me via call), I'd write down somewhere where this was sourced from.

src/metrics/kubernetes.py
```diff
@@ -0,0 +1,97 @@
...
+        pods = self._api.list_namespaced_pod(
+            namespace, label_selector=",".join(label_selector)
+        )
+        if not pods.items:
+            raise RuntimeError("Unable to find service pods.")
+        name = pods.items[0].metadata.name
+
+        return pods, name
+
+    def _find_service_port_name_in_pods(self, port, pods) -> int:
```
AlejandroCabeza, line 71: If you can, add parameter typing where it's missing.

src/metrics/kubernetes.py
```diff
@@ -0,0 +1,97 @@
...
+        if isinstance(port, str):
+            for container in pods.items[0].spec.containers:
+                for container_port in container.ports:
+                    if container_port.name == port:
+                        port = container_port.container_port
+                        break
```
AlejandroCabeza, line 77: I believe you could replace this for a straight
```py
return container_port.container_port
```
Bear in mind, though, due to the nature of this function the return type is not `int` but `Optional[int]`.

AlejandroCabeza (approved):
LGTM

src/metrics/scrapper.py
AlbertoSoutullo: Yup, thought about that as well. Will add it.

src/metrics/kubernetes.py
AlbertoSoutullo: Will put in as a link as well, it is the same one.

src/metrics/kubernetes.py
AlbertoSoutullo: Thanks, I missed that.

src/metrics/kubernetes.py
AlbertoSoutullo: As it is right now, it is no longer an `Optional`, right?

AlbertoSoutullo: True, that line is unnecessary. Removed.

PR #16, opened 2024-03-22
Alberto/data manager

### LINKED ISSUE #11: Move all data management to specific class
All functionality related to data management as DataFrames should be extracted to its own class.
Currently, both **Scrapper** and **Plotter** class are doing data conversions when it's not their responsibility.

AlbertoSoutullo:
Responsibilities of data management (create, merge, and other DF operations) is moved to its correspondent classes.

Closes #11

### DIFF
--- a/src/data/data_file_handler.py
+++ b/src/data/data_file_handler.py
@@ -39,7 +39,7 @@ class DataFileHandler(DataHandler):
                     logger.error(msg)
 
     def add_dataframe_from_file_as_mean(self, file_path: Path) -> Result[str, str]:
-        if file_utils.check_if_path_exists(file_path):
+        if file_path.exists():
             self._dump_mean_df(file_path)
             return Ok(f"{file_path} dumped to memory.")
 
@@ -54,4 +54,3 @@ class DataFileHandler(DataHandler):
         df_mean = df.mean()
         df_mean = pd.DataFrame(df_mean, columns=[file_name])
         self._dataframe = pd.concat([self._dataframe, df_mean], axis=1)
-
--- a/src/data/data_file_handler.py
+++ b/src/data/data_file_handler.py
@@ -44,9 +44,6 @@ class DataFileHandler(DataHandler):
 
         return Err(f"{file_path} cannot be dumped to memory.")
 
-    def get_dataframe(self) -> pd.DataFrame:
-        return self._dataframe
-
     def _dump_mean_df(self, file_path: Path):
         df = pd.read_csv(file_path, parse_dates=['Time'], index_col='Time')
         file_name = file_utils.get_file_name_from_path(file_path)
--- a/src/data/data_file_handler.py
+++ b/src/data/data_file_handler.py
@@ -1,0 +1,44 @@
+# Python Imports
+import logging
+import pandas as pd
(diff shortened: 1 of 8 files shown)


src/utils/file_utils.py
```diff
@@ -16,10 +21,19 @@ def read_yaml_file(file_path: str) -> Dict:
...
+    if check_if_path_exists(path):
+        files = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+        logger.info(f"Found {len(files)} in {path}")
+        logger.debug(f"Files are: {files}")
+        return Ok(files)
+
+    return Err(f"{path} does not exist.")
+
 
+def get_file_name_from_path(file_path: Path) -> str:
```
AlejandroCabeza, line 34: I think the usages of this function might've been removed. Also, does it provide any value?

src/utils/file_utils.py
```diff
@@ -16,10 +21,19 @@ def read_yaml_file(file_path: str) -> Dict:
     return data
 
 
-def get_files_from_folder_path(path: str) -> List:
-    return [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+def get_files_from_folder_path(path: Path) -> Result[List, str]:
+    if check_if_path_exists(path):
+        files = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
```
AlejandroCabeza, line 26: Alternative if you wanna stick to `Path`:
```suggestion
if path.exists():
    files = [p for p in path.glob("*") if p.is_file()]
```

AlejandroCabeza (approved):

src/data/data_file_handler.py
```diff
@@ -0,0 +1,57 @@
...
+            match self.add_dataframe_from_file_as_mean(location / file_path):
+                case Ok(msg):
+                    logger.info(msg)
+                case Err(msg):
+                    logger.error(msg)
+
+    def add_dataframe_from_file_as_mean(self, file_path: Path) -> Result[str, str]:
+        if file_utils.check_if_path_exists(file_path):
+            self._dump_mean_df(file_path)
+            return Ok(f"{file_path} dumped to memory.")
```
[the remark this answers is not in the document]
AlbertoSoutullo: Yep, between all the options that we could do here, I guess returning the object is the “better” one. It would be nice to return something like `_`, but we can't, and `None` feels weird.

src/data/data_file_handler.py
```diff
@@ -0,0 +1,57 @@
...
+
+        return Err(f"{file_path} cannot be dumped to memory.")
+
+    def get_dataframe(self) -> pd.DataFrame:
+        return self._dataframe
```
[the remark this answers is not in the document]
AlbertoSoutullo: Yup, forget about that. I am too used to the “standard” get methods.
Also, I moved this to the parent class, because I think it makes much more sense.

src/utils/file_utils.py
AlbertoSoutullo: It is still being used, but it is true that it does not provide much value... Will remove it.

AlejandroCabeza (approved):
LGTM

Issue #7, opened 2024-03-07
Yaml file improvements


AlbertoSoutullo:
- [x] CSV file naming, file dumping location...
- [ ] Add more plotting config (plotname, labels, ...)

PR #17, opened 2024-03-26
Scrape yaml improvements

AlbertoSoutullo:
One part of issue #7 

Depends on PR #16

### DIFF
--- a/src/utils/path.py
+++ b/src/utils/path.py
@@ -1,14 +1,46 @@
 # Python Imports
+import os
+import logging
 from pathlib import Path
+from typing import List
 from result import Result, Err, Ok
 
 
-def prepare_path(folder_location: str, file_name: str) -> Result[Path, str]:
-    output_dir = Path(folder_location)
+logger = logging.getLogger(__name__)
+
+
+def prepare_path(file_location: str) -> Result[Path, str]:
+    file_location_path = Path(file_location)
+    dirname = get_parent_folders_from_file_path(file_location_path)
+
+    if file_location_path.exists():
+        logger.warning(f'{file_location_path} is already existing. File will be overwritten.')
 
     try:
-        output_dir.mkdir(parents=True, exist_ok=True)
+        Path(dirname).mkdir(parents=True, exist_ok=True)
     except OSError as e:
-        return Err(f'Error creating {output_dir}. {e}')
+        return Err(f'Error creating {dirname}. {e}')
 
-    return Ok(output_dir / file_name)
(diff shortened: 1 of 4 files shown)


src/utils/path.py
```diff
@@ -1,14 +1,46 @@
...
 
 
-def prepare_path(folder_location: str, file_name: str) -> Result[Path, str]:
-    output_dir = Path(folder_location)
+logger = logging.getLogger(__name__)
+
+
+def prepare_path(file_location: str) -> Result[Path, str]:
+    file_location_path = Path(file_location)
+    dirname = get_parent_folders_from_file_path(file_location_path)
```
AlejandroCabeza, line 14: `Path` to the rescue 😆
```suggestion
dirname = file_location_path.parent
```
Which given you wanna return `file_location_path`, means that instead of this assigment, you can just replace the line 20 for
```py
file_location_path.parent.mkdir(parents=True, exist_ok=True)
```
And kill 2 birds with one stone

src/utils/path.py
```diff
@@ -1,14 +1,46 @@
...
+
+
+def get_files_from_folder_path(path: Path) -> Result[List, str]:
+    if check_if_path_exists(path):
+        files = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+        logger.info(f"Found {len(files)} in {path}")
+        logger.debug(f"Files are: {files}")
+        return Ok(files)
+
+    return Err(f"{path} does not exist.")
```
AlejandroCabeza, line 34: This function has reverted

src/utils/path.py
```diff
@@ -1,14 +1,46 @@
...
+    return file_path.name
+
+
+def get_parent_folders_from_file_path(file_path: Path) -> str:
+    return os.path.dirname(file_path)
+
 
-    return Ok(output_dir / file_name)
+def check_if_path_exists(path: Path) -> bool:
+    return os.path.exists(path)
```
AlejandroCabeza, line 46: Some of this functions were removed/edited in the previous PR, and those changes have been reverted.

src/utils/path.py
AlbertoSoutullo: Hehe

src/utils/path.py
AlbertoSoutullo: Yes indeed

src/utils/path.py
AlbertoSoutullo: Yup, I wrongly added something in the interactive rebase hehe

PR #19, opened 2024-03-29
Update plotter and data classes tests

### LINKED ISSUE #6: Plotting module tests

AlbertoSoutullo:
Created tests for new Data classes, and also updated tests for plotter class as they are very related.

Fix tests for metrics module will be in a separated PR

Closes #6


src/data/tests/test_data_file_handler.py
```diff
@@ -0,0 +1,68 @@
...
+from unittest.mock import patch
+
+# Project Imports
+from src.data.data_file_handler import DataFileHandler
+
+
+class TestDataFileHandler(unittest.TestCase):
+
+    def setUp(self):
+        self.current_directory = os.path.dirname(os.path.abspath(__file__))
```
AlejandroCabeza, line 17: Microimprovement (maybe): If this `self.current_directory` is used as a constant, that is, its value won't change throughout the tests, you can instead initialise it in
```py
@classmethod
def setUpClass(cls):
    cls.current_directory = ...
```
This method is only called once, on class initialisation, contrary to `setUp` which is called once for every case.

src/data/tests/test_folder_1/test1.csv
AlejandroCabeza, line 1: I'd suggest grouping the files containing sample data for tests under a common directory with an identifiable name. E.g.: `tests/resources/...`, `tests/data/...`, etc.

src/plotting/plotter.py
```diff
@@ -37,16 +36,12 @@ def _create_plot(self, plot_name: str, plot_specs: Dict):
...
-                case Ok(msg):
-                    logger.info(msg)
-
-                    subplot_df = file_data_hanlder.dataframe
-                    subplot_df = DataHandler.prepare_dataframe_for_boxplot(subplot_df)
-                    self._add_subplot_df_to_axs(subplot_df, i, axs, plot_specs)
-                case Err(msg):
-                    logger.error(msg)
+            file_data_handler = DataFileHandler()
+            file_data_handler.add_dataframes_from_folders_as_mean(subplot_path_group)
```
AlejandroCabeza, line 40: If this function errors out (logging the Err and then returning), doesn't it impact the continuity of this function?

src/plotting/tests/test_plotter.py
```diff
@@ -28,37 +28,55 @@ def test__create_plot(self):
...
-        plotter._create_subplot_paths_group = MagicMock()
-        plotter._insert_data_in_axs(["test", "test"], None)
+        plotter._insert_data_in_axs([["path1"], ["path2"]], [0], {"data": ["data1"]})
 
-        self.assertEqual(plotter._create_subplot_df.call_count, 2)
+        self.assertEqual(mock_data_file_handler.call_count, 2)
+        self.assertEqual(mock_data_handler.call_count, 2)
         self.assertEqual(plotter._add_subplot_df_to_axs.call_count, 2)
 
     def test__save_plot(self):
```
AlejandroCabeza, line 42: Does this function save the plot to a local file? Is it feasible to actually save that file, and test whether it exists?

src/data/tests/test_data_file_handler.py
AlbertoSoutullo: For some reason, I thought what I had was doing that per class, not per test. Ty!

src/data/tests/test_folder_1/test1.csv
AlbertoSoutullo: I like that, changing it.

src/plotting/plotter.py
AlbertoSoutullo: Hmm, what do you mean? This function doesn't return if it has an error. It writes the error, but keep's reading folders. If it cannot read data, it will create an empty dataframe. And if it is missing data, that will be logged, and the dataframe will be created with missing data.

src/plotting/tests/test_plotter.py
AlbertoSoutullo: The function itself yes, the test function no.
The function itself just does 3 calls from matplotlib:

- tight_layout
- savefig (which saves the plot to a local file)
- show (open the plot in a new window)
I don't think it is in the scope of this test to check if the file is created or not, since that should be already tested by matplotlib.

AlejandroCabeza: Right, but instead of doing the 3 mock checks, wouldn't it be simpler to just run the function without mocking, and check whether the file has been generated? Or does "creating the file" not imply those 3 listed calls?

src/plotting/plotter.py
AlejandroCabeza: I'm wondering whether the `add_dataframes_from_folders_as_mean` function should be doing all the "if it's ok then do stuff, it it's err then log the error" work, or just return a `Result` (as in, not handling the `Err` case) and let callers handle it.

src/plotting/tests/test_plotter.py
AlbertoSoutullo: The only call generating the file is the second one. And to check if the file is created doesn't prove that it worked as expected. In order to check that the function works as expected, we would need to:
- Add specific data to a dataframe
- Plot it
- Save the plot as a `resources` for the test
- Each time we run the test, we compare the generated file with the previously saved file

And again, I don't think it is our responsibility to test if the `savefig` method works at that level.

src/plotting/plotter.py
AlbertoSoutullo: There is a specific reason about why I did it like this. Let me elaborate:
In the configuration file, you can set up a plot like the following example:
- I want to get the metric `bandwidth` from the waku versions `v0.24`, `v0.25` and `v0.26`.
In order to display this data in the same plot, what `add_dataframes_from_folders_as_mean` will receive as a List will be `['v0.24/bandwidth', 'v0.25/bandwidth', 'v0.26/bandwidth']`.

Each of those folders will be a subplot inside the same plot. If `v0.24/bandwidth` doesn't exist for any reason, the plot will still be created with `0.v25` and `0.v26`. So returning a `Result` here might be misleading. Because if a folder is missing, it can be argued that might be `Ok` or `Err`.

src/plotting/tests/test_plotter.py
AlejandroCabeza: Oh, no, absolutely, we are not supposed to test if `savefig` works. I was just think, perhaps the `savefig` method is called if (and only if) all the others work as expected. Then the assertion could be as easy as that. Got it.

Well, then just a future reference (feel free to use, or don't) which is this [`mock.assert_called_once_with`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with).

AlbertoSoutullo: Yes, sometime I use that, sometimes I use the call_count to 1. Depending on which one is closer to copy-paste heh.

AlejandroCabeza (approved):
lgtm

PR #21, opened 2024-04-03
Download pod logs

AlbertoSoutullo:
`kubernetes.config.` is something that is not pickable by python, but we need this configuration to access to the API. 
I could do several things here, but most of them were about taking out the configuration and hence the API access from the class.
This was a very specific case where it is one of the very few occasions we will want to do something asynchronously in the cluster. I decided to create a static method in the class, so we don't have to take out the client and its configuration outside the class. This allows us to do calls asynchronously, since each call will create a new client configuration instead of trying to pickle it.

### DIFF
--- a/src/metrics/kubernetes.py
+++ b/src/metrics/kubernetes.py
@@ -1,17 +1,22 @@
 # Python Imports
 import socket
 import logging
+import kubernetes
+import multiprocessing
 from typing import List, Tuple
-from kubernetes.client import CoreV1Api, V1PodList, V1Service
+from kubernetes.client import V1PodList, V1Service
 from kubernetes.stream import portforward
 
+from src.utils import path
 
 logger = logging.getLogger(__name__)
 
 
 class KubernetesManager:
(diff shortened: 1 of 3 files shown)


src/metrics/kubernetes.py
```diff
@@ -31,6 +37,39 @@ def create_connection(self, address, *args, **kwargs) -> socket.socket:
 
         return pf.socket(port)
 
+    @staticmethod
+    def download_logs_from_pod_asyncable(kube_config: str, namespace: str, pod_name: str,
+                                         location: str):
+        kube_client = kubernetes.config.new_client_from_config(kube_config)
```
AlejandroCabeza, line 43: Why not make the function a `method` and use `self._kube_client`?

AlbertoSoutullo: Discussed in the offsite. If you have any other idea, let me know. For now, I think this is the “best” I have in mind.

AlejandroCabeza (approved):

PR #29, opened 2024-05-24
Update yamls

AlbertoSoutullo:
Rebased midstrap branch onto master. 

Moved previous accelerated rubi folder files onto accelerated deployment.

Please, let me know if anything else should be here.

After this squashed merge, `zorling/midstrap` will be deleted. Start future changes from `master`, and if they are stable, PR them so the rebase should be easier and quicker c:

Note that multus scripts, kubernetes scripts and other stuff is removed as afaik they were not being used in this repo.
If something that is necessary is missing, let me know and I will recover it.


Zorlin:
LGTM

PR #30, opened 2024-05-24
Publisher changes

AlbertoSoutullo:
- Changed permanent injection to a given number of messages
- Handle exceptions when injecting messages. Injection errors were not being handled properly.
- Correctly sent KBytes (`/1000` instead `/1024`)
- Added a number to each message, so it is easier to track in the log.
- Changed prints to better traceability

The idea with this is to inject a given number of messages. With this, we can start a simulation, inject the messages, and finish the simulation in a cleaner manner.

Right now we "cut" the simulation, that makes the results a bit skewed depending on the current state of the mesh and the messages. 

So now if we want a simulation of 20 minutes, we can put `1msg/s` and `72000` messages. This will also help us getting better information at reliability level.

### DIFF
--- a/accelerated_deployment/publisher_headless/traffic.py
+++ b/accelerated_deployment/publisher_headless/traffic.py
@@ -6,7 +6,7 @@ import asyncio
 import aiohttp
 import argparse
 import socket
-from itertools import cycle
+
 
 async def check_dns_time(node: str) -> str:
     start_time = time.time()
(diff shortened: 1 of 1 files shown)


AlbertoSoutullo:
Also wanted to add that another benefit that this change has is that we can:
- Start a sim, give it 7~ minutes to stabilize.
- Inject N messages.
- Kill the simulation once the publisher finishes (maybe give it a minute or 2 after it finishes).

Then we can script this in a batch of simulations, let the lab run in the night, and we can get the results in the morning, as we are not using more than 3k nodes and 1msg/s.
I always prefer to kill and restart simulations to have fresh Waku nodes, instead of expanding the size. In this case, we are sure that all the simulations are independent, and we are not adding an extra layer of problems in the middle.

This feels a little bit more natural that inject in a while true, and suddenly kill it in the middle to start a new one.

Zorlin:
> Also wanted to add that another benefit that this change has is that we can:
> 
> * Start a sim, give it 7~ minutes to stabilize.
> * Inject N messages.
> * Kill the simulation once the publisher finishes (maybe give it a minute or 2 after it finishes).
> 
> Then we can script this in a batch of simulations, let the lab run in the night, and we can get the results in the morning, as we are not using more than 3k nodes and 1msg/s. I always prefer to kill and restart simulations to have fresh Waku nodes, instead of expanding the size. In this case, we are sure that all the simulations are independent, and we are not adding an extra layer of problems in the middle.
> 
> This feels a little bit more natural that inject in a while true, and suddenly kill it in the middle to start a new one.

LGTM. I didn't have any objection anyways, just hadn't gotten to it yet! Thanks for pushing things forwards, this is great work.

Zorlin (approved):
LGTM

PR #31, opened 2024-05-28
Nim-libp2p deploy

AlbertoSoutullo:
I introduced some small change in the [dst-test-node](https://github.com/vacp2p/dst-gossipsub-test-node) repo.
In order to help the p2p team with the monthly releases, I created a couple of yaml files to do the deployment.

If you remember in my initial deployment, I had one service per node. I [changed the code](https://github.com/vacp2p/dst-gossipsub-test-node/blob/dockerized/main.nim#L158) so now instead of doing this, they will use a single service for doing the random connections.

I set this to a different namespace `zerotesting-nimlibp2p`. Pods will use port `5000` for communication, and port `8000` as metric endpoint.

`command` values are: <minute> <hour> of the cronjob to start all nodes at once.

It is a simple approach right now, but Ideally after receiving feedback from the p2p team we will improve the deployment and also maybe doing changes to the dst-test-node image.

When you have a moment, please take a look at the yamls to double-check that everything is correct. 
Also, I guess this new port should be added somewhere in the pod-monitor service of the lab?


Zorlin:
Hi Alberto,

Yes, I believe this looks good.

I would prefer slightly that we break out the change to remove the old deploy/ folder into its own PR for cleanliness, but I don't mind if we merge this as is and note that for next time

Looks good to me overall. :+1: LGTM

> Also, I guess this new port should be added somewhere in the pod-monitor service of the lab?
Yup. Please coordinate with me to make the change to the Rancher Monitoring Helm chart's `values.yaml` file to fully implement this change. I would suggest I document what needs to be done and then either you or the Technician applies the fix.

Zorlin (approved):
Suggestion, but we can ship without it -

* Rename accelerated_deployment to just deployment and delete the old deployment folder

AlbertoSoutullo:
> Suggestion, but we can ship without it -
> 
> * Rename accelerated_deployment to just deployment and delete the old deployment folder

Yup, the old folder was deleted, I just renamed the current one with the last push. Ty.

PR #32, opened 2024-06-06
Ignore columns option for plotting

AlbertoSoutullo:
This PR is a bit coupled, sorry beforehand.

In one side, I added an argument for ignore columns that start with a certain string.
On the other side, while doing that, I refactored a bit the methods for reading from files since they were a bit coupled. Also, renamed `add` methods for `concat`, since it explains better what they are doing on the inside.

If it is too hard to understand let me know, and I will try to split it in two PRs.

### DIFF
--- a/src/data/data_file_handler.py
+++ b/src/data/data_file_handler.py
@@ -2,7 +2,7 @@
 import logging
 import pandas as pd
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 from result import Result, Ok, Err
 
 # Project Imports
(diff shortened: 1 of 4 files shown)


src/data/data_file_handler.py
```diff
@@ -13,27 +13,26 @@
...
+                                                           points)
                     folder_df["class"] = f"{folder_path.parent.name}/{folder_path.name}"
                     self._dataframe = pd.concat([self._dataframe, folder_df])
                 case Err(error):
                     logger.error(error)
 
-    def _add_files_as_mean(self, target_df: pd.DataFrame, data_files_path: List,
-                           location: Path, points: int) -> pd.DataFrame:
+    def _concat_files_as_mean(self, target_df: pd.DataFrame, data_files_path: List,
+                              location: Path, points: int) -> pd.DataFrame:
```
AlejandroCabeza, line 33: If, for some reason, there's no match, the function will return Option[pd.DataFrame], I'd update that just to the on the safe side.
That'd also happen if `data_files_path` is empty.

src/data/data_file_handler.py
```diff
@@ -13,27 +13,26 @@
...
         for file_path in data_files_path:
-            match self.add_data_from_file(target_df, location / file_path, points):
+            match self._concat_data_as_mean_from_file(target_df, location / file_path, points):
                 case Ok(result_df):
                     logger.info(f"{file_path} added")
                     target_df = result_df
```
AlejandroCabeza, line 38: On the other hand, can the `target_df = result_df` be changed to a `return result_df`? Or do we want to only return the last value assigned?

src/data/data_file_handler.py
```diff
@@ -13,27 +13,26 @@
 
 
 class DataFileHandler(DataHandler):
-    def __init__(self):
-        super().__init__()
-        self._dataframe = pd.DataFrame()
+    def __init__(self, ignore: Optional[List] = None):
+        super().__init__(ignore)
```
AlejandroCabeza, line 17: Now it's not providing any value anymore, you can remove it.

src/data/data_file_handler.py
AlbertoSoutullo: Hmm I am not sure if I am following you. `target_df` is a dataframe that you provide, to which the data will be concatenated. If for some reason any data was concatenated, you still will get the original `target_df`. So there is no option to return a `None`. Also, you mean `Optional` right?

src/data/data_file_handler.py
AlbertoSoutullo: I also don't follow you here.
If we change `target_df = result_df` to a `return result_df` the for loop will not be completed, and we will return only the first `file_path` appended.
This is not returning the last value assigned, it is appending all values from all `data_files_path` into `target_df`.

AlejandroCabeza: Answered in the other comment.

src/data/data_file_handler.py
AlejandroCabeza: Ah, right, got it. I hadn't properly understood the function, and you just made it clear. Thanks :)

P.D.: Yes, I meant `Optional`.

AlejandroCabeza (approved):

PR #43, opened 2024-10-25
Filter lightpush 

AlbertoSoutullo:
Adapted code to analyze filter-lightpush simulations.
Crated docker containers for address retrieval and get filter messages
Improved message injector
Improved get ENR container

### DIFF
--- a/deployment/docker/get_addres/getaddress.sh
+++ b/deployment/docker/get_addres/getaddress.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Number of addrs to process, default to 3 if not specified
+# Number of addrs to process, default to 1 if not specified
 num_addrs=${1:-1}
 
 # Service name to query, default to "zerotesting-bootstrap.zerotesting" if not specified
--- a/deployment/docker/get_addres/getaddress.sh
+++ b/deployment/docker/get_addres/getaddress.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Number of addrs to process, default to 3 if not specified
+num_addrs=${1:-1}
+
+# Service name to query, default to "zerotesting-bootstrap.zerotesting" if not specified
+service_name=${2:-zerotesting-bootstrap.zerotesting}
+
+# Find the IPv4 IPs of "zerotesting-bootstrap.zerotesting" using nslookup
+readarray -t pod_ips < <(nslookup "$service_name" | awk '/^Address: / { print $2 }' | head -n "$num_addrs")
+
+# Prepare the directory for addrs data
+mkdir -p /etc/addrs
+addrs_file="/etc/addrs/addrs.env"
+> "$addrs_file" # Clear the file to start fresh
+
+# Function to validate addrs
+validate_addrs() {
+    if [[ $1 =~ ^/ip ]]; then
+        return 0 # Valid
+    else
+        return 1 # Invalid
+    fi
+}
+
+# Counter for valid addrs
+valid_addrs_count=0
+
+# Get and validate the addrs data from up to the specified number of IPs
+for pod_ip in "${pod_ips[@]}"; do
+    echo "Querying IP: $pod_ip"
+    addrs=$(curl -X GET "http://$pod_ip:8645/debug/v1/info" -H "accept: application/json" | sed -n 's/.*"listenAddresses":\["\([^"]*\)".*/\1/p')
+
+    # Validate the addrs
+    validate_addrs "$addrs"
+    if [ $? -eq 0 ]; then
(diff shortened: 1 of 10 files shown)


deployment/docker/get_addres/getaddress.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# Number of addrs to process, default to 3 if not specified
+num_addrs=${1:-1}
```
AlejandroCabeza, line 4: Doesn't match the comment. Default is `1`.

deployment/docker/get_filter_messages/filter_msg_retriever.py
```diff
@@ -0,0 +1,91 @@
...
+
+    return parser.parse_args()
+
+
+def fetch_all_messages(base_url: str, headers: Dict, address: str) -> Optional[List[str]]:
+    response = requests.get(base_url, headers=headers)
+    if response.status_code != 200:
+        logging.error(f"Error fetching data: {response.status_code}")
+        logging.error(response.text)
+        return None
```
AlejandroCabeza, line 38: Double checking with you if it'd make more sense to use `Result` rather than `Optional`.

deployment/docker/get_filter_messages/filter_msg_retriever.py
```diff
@@ -0,0 +1,91 @@
...
+
+    data = response.json()
+    messages = [message_data['payload'] for message_data in data]
```
AlejandroCabeza, line 41: Microoptimization, in case you wanna go for it (and fits, ofc). Generally it's better to rely on generators unless you need to read the list multiple times.

deployment/docker/get_filter_messages/filter_msg_retriever.py
```diff
@@ -0,0 +1,91 @@
...
+            if result:
+                all_messages.extend(result)
+
+    if len(all_messages) > 1:
+        it = iter(all_messages)
+        len_messages = len(next(it))
+        if not all(len(_) == len_messages for _ in it):
+            print("False")
+        else:
+            print("True")
```
AlejandroCabeza, line 83: I'm a bit confused with the intention of this, could you explain it to me?

Also, I'd suggest only using `_` when the variable is to be "discarded".

deployment/publisher_headless/traffic.py
```diff
@@ -6,92 +7,137 @@
 import aiohttp
 import argparse
 import socket
+import logging
 
+logging.basicConfig(level=logging.INFO)
 
-async def check_dns_time(node: str) -> str:
+
+async def check_dns_time(service: str) -> (str, str):
```
AlejandroCabeza, line 15: A useful tool in these kind of scenarios are [namedtuples](https://docs.python.org/3/library/collections.html#collections.namedtuple): Just a tuple but with named attributes. You can use the name as accessor for the attributes, and they also serve as documentation. Just letting you know in case you weren't aware of them and it makes sense to use them here (or in the future).

AlejandroCabeza (approved):

deployment/docker/get_addres/getaddress.sh
AlbertoSoutullo: God bless copy paste

deployment/docker/get_filter_messages/filter_msg_retriever.py
AlbertoSoutullo: It is a script on a standalone docker image, I would prefer to keep it as lightweight and simple as possible. 
Also, the functionality will remain rather small. We do not really benefit from `Result`. If it was in the core of the project, I would definetly use `Result`.

deployment/docker/get_filter_messages/filter_msg_retriever.py
AlbertoSoutullo: First thing I want from it is the `len`, so I don't think it would fit in this case.

deployment/docker/get_filter_messages/filter_msg_retriever.py
AlbertoSoutullo: `all_messages` is a `List[List]`.
What I want to know if all the lists have the same length. So I get the iterator, I check the first `len` to get the first number as a baseline, and then I check if all `len` from the remaining sublists is the same.
If it is not, then we have False, as not all lists have the same `len`.
I use `_` because the variable is indeed discarded.

deployment/publisher_headless/traffic.py
AlbertoSoutullo: Yes! But again, as this is another script in an isolated image, didn't want to "boilerplate" too much the code.

PR #52, opened 2025-01-30
Cleaning and documenting repo

AlbertoSoutullo:
- Removed old files no longer used
- Add READMEs for utilities we do for simulations (containers)
- Regarding this, imo we could start from version 1 from now on, and always document which version we use for which experiment.

I will create another PR with the files I have been using for running simulations.


AlbertoSoutullo:
> > Regarding this, imo we could start from version 1 from now on, and always document which version we use for which experiment.
> 
> Can we start at v0.1.0 instead of v1.0.0 except for utilities that have "Graduated" and are considered stable and unchanging?
> 
> https://semver.org

I was thinking about that, as I have `v0.12` versions already of publisher, for example.

Stable in this context is tricky because it depends on the modifications that Waku or other projects will do. Same for others. Most of them were used for the reports, so I would consider them "stable" until we need to expand them depending on what we want to do. To be honest, I don't really mind going with `v0.x` but we should agree on that.

Zorlin:
> > > Regarding this, imo we could start from version 1 from now on, and always document which version we use for which experiment.
> > 
> > 
> > Can we start at v0.1.0 instead of v1.0.0 except for utilities that have "Graduated" and are considered stable and unchanging?
> > https://semver.org
> 
> I was thinking about that, as I have `v0.12` versions already of publisher, for example.
> 
> Stable in this context is tricky because it depends on the modifications that Waku or other projects will do. Same for others. Most of them were used for the reports, so I would consider them "stable" until we need to expand them depending on what we want to do. To be honest, I don't really mind going with `v0.x` but we should agree on that.

Good point - in that case I don't mind either way, but we should make sure we specify which way we go. :)

AlbertoSoutullo:
> > I was thinking about that, as I have `v0.12` versions already of publisher, for example
> > Stable in this context is tricky because it depends on the modifications that Waku or other projects will do. Same for others. Most of them were used for the reports, so I would consider them "stable" until we need to expand them depending on what we want to do. To be honest, I don't really mind going with `v0.x` but we should agree on that.
> 
> Good point - in that case I don't mind either way, but we should make sure we specify which way we go. :)

I would go with this, as we have basically everything for a baseline. And then making sure we track correctly the versions, we report which one we are using in the experiments, and a changelog also.

Zorlin:
As for the PR - LGTM but I think it's time to let Midstrap go ;) We're unlikely to use midstrap nodes again.

Zorlin (changes requested):
Looks good, but let's get rid of midstrap files too

Zorlin:
> > I will create another PR with the files I have been using for running simulations.
> 
> Yes! This is what I mentioned in the first comment, I will create a separated PR for that.

Ah okay! My mistake, sorry for missing that. Will approve! :)

Zorlin (approved):
Approved! LGTM

PR #54, opened 2025-01-31
Deployment utils

AlbertoSoutullo:
Added all files that I have been using to run experiments.
Deleted old ones.


AlbertoSoutullo:
Ideally we would have code for setting up those yamls, but we need more hands to do all at once. So for now, I would save them to re-use when we need to repeat the experiment.

PR #58, opened 2025-02-13
Scale value in plot function util

AlbertoSoutullo:
QoL PR to scale the value of the numbers in the plot, to match axis if needed.


Zorlin (approved):
LGTM

PR #66, opened 2025-05-23
WIP: Analysis refactor

AlbertoSoutullo:
Refactor of the structure. With this, I intend to:
1. Decouple code that is entangled, ie: Waku-Victoria, as it was thought for DST experiments only.
2. Extend modularity, allowing to generate different analysis depending on the monitoring stack that is being used, ie:
  - Vaclab + Victoria
  - Wakufleet + elastic
3. Make it easy to change the analysis if the stack is changed anywhere, ie: we set up a different monitoring stack in Vaclab.
4. Allow us to work "seamlessly" of where the data is generated. We only need to implement the stack workflow, the helper classes (Readers, Tracers, Analysis) are modular and should remain as closed as possible.
5. Introduce new project analysis (Nomos) without interfering with other analysis.
6. Prepare the path for the upcoming Waku log parsel tooling task.


PearsonWhite (approved):
Discussed in a call.

PR #70, opened 2025-06-04
Adapt experiments to python scaffold

PearsonWhite:


experiments/deployment/common.py
```diff
@@ -0,0 +1,115 @@
...
+
+from kubernetes.client import ApiClient
+from pydantic import BaseModel, PositiveInt
+from ruamel import yaml
+
+from kube_utils import (
+    maybe_dir,
+    poll_namespace_has_objects,
+    wait_for_no_objs_in_namespace,
+)
```
AlbertoSoutullo, line 20: Mentioned in general comment

experiments/deployment/common.py
```diff
@@ -0,0 +1,115 @@
...
+
+logger = logging.getLogger(__name__)
```
AlbertoSoutullo, line 22: Mentioned in general comment

experiments/deployment/common.py
```diff
@@ -0,0 +1,115 @@
...
+class BaseExperiment(ABC, BaseModel):
+    """Base experiment that add an ExitStack with `workdir` to `run` and uses an internal `_run`.
+
+    How to use:
+        - Inherit from this class.
+        - Call `BaseExperiment.add_args` in the child class's `add_parser`
+        - Implement `_run` in the child class.
+    """
+
+    def add_args(subparser):
```
AlbertoSoutullo, line 34: If this is a static method, let's use the @staticmethod decorator.

experiments/deployment/common.py
```diff
@@ -0,0 +1,115 @@
...
+            test.run(values, workdir)
+            time.sleep(600)
+
+
+# Example usage:
+# config.load_kube_config(config_file="./kube_config.yaml")
+# client = ApiClient()
+# run_waku_regression_nodes(
+#     "./workdir", client, "./values.yaml", [100, 200, 300], [1, 5, 10]
+# )
```
AlbertoSoutullo, line 115: What is the purpose of this?

experiments/deployment/nimlibp2p/experiments/regression.py
```diff
@@ -37,9 +36,24 @@
 
 class NimRegressionNodes(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True)
-    api_client: ApiClient = Field(default=client.ApiClient())
     release_name: str = Field(default="nim-regression-nodes")
 
+    def add_args(subparser):
```
AlbertoSoutullo, line 41: Same, static decorator.

experiments/deployment/nimlibp2p/experiments/regression.py
```diff
@@ -37,9 +36,24 @@
...
+            required=False,
+            help="For nimlibp2p tests only. The delay before nodes activate in string format (eg. 1hr20min)",
+        )
+
+    def add_parser(subparsers):
+        subparser = subparsers.add_parser(
+            "nimlibp2p-regression-nodes", help="Run a regression_nodes test using waku."
+        )
+        BaseExperiment.common_flags(subparser)
+        NimRegressionNodes.add_args(subparser)
```
AlbertoSoutullo, line 55: Decorator and typing.

experiments/deployment/regression_nodes.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def add_parser(subparsers):
+        regression_nodes = subparsers.add_parser(
+            "regression-nodes", help="Run a regression_nodes test."
+        )
+        regression_nodes.add_argument(
+            "--type", type=str, choices=["waku", "nim"], required=True, help=""
+        )
+        BaseExperiment.add_args(regression_nodes)
+        NimRegressionNodes.add_args(regression_nodes)
```
AlbertoSoutullo, line 29: Decorator and Typing

experiments/deployment/regression_nodes.py
```diff
@@ -0,0 +1,41 @@
...
+        logger.debug(f"args: {args}")
+
+        if args.type == "waku":
+            experiment = WakuRegressionNodes()
+            experiment.run(api_client=api_client, args=args, values_yaml=values_yaml)
+        elif args.type == "nim":
+            experiment = NimRegressionNodes()
+            experiment.run(api_client=api_client, args=args, values_yaml=values_yaml)
+        else:
+            raise ValueError(f"Unknown regression experiment type: `{args['type']}`")
```
AlbertoSoutullo, line 41: If all classes will implement `run`, we should call it abstractly here, and if extra params are needed, we can just use kwargs.

experiments/deployment/waku/builders.py
```diff
@@ -0,0 +1,125 @@
...
+
+    deployment_dir: str = Field(default=os.path.dirname(__file__))
+
+    def _get_excluded_yamls(self, work_sub_dir, service):
+        return [
+            path
+            for path in glob.glob(
+                os.path.join(work_sub_dir, "templates", "*.yaml"), recursive=False
+            )
+        ]
```
AlbertoSoutullo, line 38: This could be moved to a yamls utils .py. Also, service is not being used (?).

experiments/deployment/waku/experiments/regression.py
```diff
@@ -0,0 +1,176 @@
...
+        BaseExperiment.add_args(subparser)
+
+    def _run(
+        self,
+        api_client: ApiClient,
+        workdir: str,
+        args: argparse.Namespace,
+        values_yaml: Optional[yaml.YAMLObject],
+        stack: ExitStack,
+    ):
```
AlbertoSoutullo, line 48: I think this method could be refactored in smaller functions.
Also, there is a common behaviour of deploy and wait that could also be abstracted into a single function.

experiments/deployment/waku/experiments/regression.py
```diff
@@ -0,0 +1,176 @@
...
+            resources_to_cleanup = get_cleanup_resources(deployments)
+            logger.info(f"Resources to clean up: `{resources_to_cleanup}`")
+
+            logger.info("Start cleanup.")
+            cleanup_resources(resources_to_cleanup, namespace, api_client)
+            logger.info("Waiting for cleanup.")
+            wait_for_cleanup(resources_to_cleanup, namespace, api_client)
+            logger.info("Finished cleanup.")
+
+        return cleanup
```
AlbertoSoutullo, line 142: I guess this could be in `kube_utils`, right?

experiments/deployment/waku/experiments/regression.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def get_flag_value(self, flag: str, command: List[str]) -> int:
+        for node in command:
+            matches = re.search(f"--{flag}=(?P<numMessages>\\d+)", node)
+            try:
+                return int(matches["numMessages"])
+            except (TypeError, IndexError):
+                pass
+        return None
```
AlbertoSoutullo, line 151: Same as before, this could be used in different places. Also, missing `Optional` in return typing.

experiments/deployment/waku/experiments/regression.py
```diff
@@ -0,0 +1,176 @@
...
+            test.run(values, workdir)
+            time.sleep(600)
+
+
+# Example usage:
+# config.load_kube_config(config_file="./kube_config.yaml")
+# client = ApiClient()
+# run_waku_regression_nodes(
+#     "./workdir", client, "./values.yaml", [100, 200, 300], [1, 5, 10]
+# )
```
AlbertoSoutullo, line 176: Purpose of this?

experiments/registry.py
```diff
@@ -0,0 +1,107 @@
...
+            if all(experiment.metadata.get(key) == value for key, value in filters.items())
+        ]
+
+    def __getitem__(self, name: str) -> ExperimentInfo:
+        try:
+            return next((info for info in self._experiments if info.name == name))
+        except StopIteration as e:
+            raise KeyError(f"No experiment for name exists: `{name}`") from e
+
+    def get(self, name: str) -> ExperimentInfo:
```
AlbertoSoutullo, line 36: `Optional`

experiments/deployment/common.py
PearsonWhite: Removed. It was just left over from when I split up the file.

experiments/deployment/waku/experiments/regression.py
PearsonWhite: Moved to kube_utils.py

PearsonWhite:
https://github.com/vacp2p/10ksim/pull/70#pullrequestreview-2935249217

>Unless I am wrong, all files are modules, so we are not using #!/usr/bin/env python3. Also, this is always executed from within the venv.

Removed `#!/usr/bin/env python3`.

>It would be nice to keep the same import structure from other modules, starting with Python Imports and then Project Imports. Alphabetically. If you find cases on other files that are wrong, please open a PR changing that c:

This should be handled when running `make format` by this line in `Makefile`: `isort --profile black -l 100`.
After this is merged, I think we should move the `Makefile` up to the root dir.

>Let's use the same logger, it is already defined in src. At some point this will be included there.

Can we put this off for another PR? The logger_config.yaml has defined loggers and it doesn't fit into the current file structure.
```
loggers:
  src:
    ...
```

>I think some yaml functions can be extracted to a yamls util.

I'll follow up on this in another comment about file layout.

>Requirements.txt should go with the main one at some point.

Agreed. I didn't want to deal with potential version conflicts here, but eventually we'll have to merge this as well as some other things.

>Let's do a more strict typing. It helps a lot with reviews.

Agreed. Ideally this would be for pretty much everything. For this PR I pretty much just added typing to help myself without spending too much time on it.

experiments/deployment/waku/experiments/regression.py
PearsonWhite: I was thinking for experiments that are essentially just build-deploy-wait we can make a base class like BasicExperiment or something. I haven't quite worked out the details yet.

One problem is that aside from building, the sanity checks, and the logging, all the code in that function is the logic of the actual experiment. We could separate the experiment into phases (deploy bootstrap, deploy nodes, etc), but I think it would just end up being a bunch of functions called once. This would spread out the logic of the function, even if it might make the higher-level view of the experiment a little more readable. I'll think about it and maybe I'll come up with a better way.

experiments/deployment/regression_nodes.py
PearsonWhite: Is this what you meant? [diff link](https://github.com/vacp2p/10ksim/pull/70/commits/36f3d124e99680fe40d771a53cb478dc549916e9?diff=unified&w=0#diff-57eec5b270b7f82dc39eda0bac239c6b15e637b10a4be8341d11a6f0b9c1cba0R29)

experiments/deployment/waku/experiments/regression.py
AlbertoSoutullo: Discussed in call

PR #72, opened 2025-06-23
Adapt regression tests to new deployment scaffold

PearsonWhite:


experiments/deployment/nimlibp2p/builders.py
```diff
@@ -0,0 +1,64 @@
...
+#         if extra_values_names is None:
+#             extra_values_names = []
+#         if cli_values is None:
+#             cli_values = CommentedMap()
+
+#         service_dir = os.path.join(self.deployment_dir, "nodes")
+#         delay = args.delay if args.delay is not None else default_delay
+#         set_delay(cli_values, "hours", "minutes", delay)
+#         # expected_start_time = datetime.now(timezone.utc) + delay
+#         # return super().build(workdir, cli_values, service, )
```
AlbertoSoutullo, line 64: If this is fully commented, should not be in the PR

AlbertoSoutullo (changes requested):
Nice cleanup. Minor comment.

experiments/deployment/nimlibp2p/builders.py
PearsonWhite: Removed.

AlbertoSoutullo (approved):
Delay calculation in nimlibp2p is a bit "by eye". We definitely need to change how the node behaves and change the deployment.
Also, let's try to do more context incremental PRs from now. The amount of files and context for this was very large, and it was hard to keep track of everything. 
Still, great amount of work. This will be super useful.

experiments/deployment/nimlibp2p/experiments/regression.py
```diff
@@ -51,144 +67,88 @@ def add_parser(subparsers) -> None:
...
+            cli_values=values,
+            name=self.release_name,
+            extra_values_names=["regression.values.yaml"],
+        )
+        if delay is None:
+            deployment = build(cli_values)
+            num_nodes = deployment["spec"]["replicas"]
+            # Assume it takes ~3 seconds to bring up each node.
+            delay = (3 * 60) + (num_nodes * 3)
+            logger.info(f"Using default delay (in seconds): {delay}")
```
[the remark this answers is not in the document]
AlbertoSoutullo: Nvm, just checked.

PR #74, opened 2025-07-02
Make regression experiments self contained

PearsonWhite:
Move .values.yaml files from node/publisher/etc to experiment folder.


AlbertoSoutullo (approved):
The double `..` is a bit smelly, but at the same time I think it makes sense (I am weird). LGTM. Thanks!

AlbertoSoutullo:
Relate to #80

PR #76, opened 2025-07-13
Changed plot stat labels use 'scale-x' instead of hardcoded 0.001

PearsonWhite:


AlbertoSoutullo (changes requested):
Thanks for doing this.

We should change it in other places where `add_boxplot_stat_labels` is called.
Also, I think a `.get` with default value `1` if it is not provided should also be ok.

And as a final touch, this should also be modified in `add_boxplot_stat_labels` function, 
because the value is being multiplied instead of divided (because it was hardcoded).
So we would end up with `value/scale_by:{fmt}` instead of `value*scale_by:{fmt}`

AlbertoSoutullo (approved):
Thanks!

PR #78, opened 2025-08-05
Add events logging for waku regression experiment

PearsonWhite:
Track events in json form so we can later parse the log and extract times.

### DIFF
--- a/experiments/deployment/base_experiment.py
+++ b/experiments/deployment/base_experiment.py
@@ -1,12 +1,16 @@
+import json
 import logging
+import os
 import shutil
 from abc import ABC, abstractmethod
 from argparse import ArgumentParser, Namespace
 from contextlib import ExitStack
-from typing import Optional
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, Optional
 
 from kubernetes.client import ApiClient
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 from ruamel import yaml
 from ruamel.yaml.comments import CommentedMap
 
@@ -24,6 +28,8 @@ class BaseExperiment(ABC, BaseModel):
         - Implement `_run` in the child class.
     """
 
+    events_log_path: Path = Field(default=Path("events.log"))
+
     @staticmethod
     def add_args(subparser: ArgumentParser):
         subparser.add_argument(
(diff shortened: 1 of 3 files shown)


PearsonWhite:
> Hmmm maybe I am not getting this PR correctly. Could you elaborate more on this?

This was motivated by two things:
1. Getting the intervals that would then be used for analysis, so we can automate the deployment to analysis pipeline.
2. I was getting really tired of looking through the logs, finding the timestamps for the experiment, converting them to UTC, and opening up grafana/vmui and entering them. This has actually been really helpful with that.

Incidentally, this separates logging about what actually happens with an experiment from logging that has to do how the deployment script is running. So maybe our script has a bunch of logging about what it's doing so we can look back in case something went wrong, but it's not relevant to what the experiment is doing. For example  `INFO - Waiting: pod 'publisher' Ready='('Ready', 'False')' ready='False'...` is nice to have to see what the script is doing, but looking back through the logs to find out when experiment events happened, it's just noise.

The particulars of what events are logged was mostly based on:
1. What was needed to create the grafana/vmui time intervals for waku-regression-nodes
2. What was common/convenient. It doesn't really cost anything to log things like `cleanup_start`, `run_finished`, and we may want to search/filter by those later.

> In this code, feels like we are not waiting for the statefulset to be `Ready`, isn't it? Or maybe I am not following it correctly?

You mean going off of `publisher_deploy_start` instead of `publisher_deploy_finished`? I'm not sure why I did that, but I don't think there's much of a difference. It probably makes sense to change that.

PearsonWhite:
Adding comment from discord conversation here:

Basic overview:
1. Set the events_log path. If it's a relative path, it's treated as relative to the --workdir path.
1. Call self.log_event, which appends to the file. It has a _preprocess_event in order to add whatever else (like maybe experiment parameters or something)
2. At the end of an experiment (or anytime I guess) you can call _metadata_event (which I guess could be a class function) to read the events.log and generate a dictionary of events/info

The way `_metadata_event` ([link](https://github.com/vacp2p/10ksim/pull/78/files#diff-3abd9f42d2956e258b24fce7d3e97e7c138532851f780f4b8acde87a42ed4a48R66)) works in this case is less elegant than I wanted, but the time offsets really got messy. We have this events_dict of all the events we want to get, how we map them, along with a timedelta to add to each time. So where we see wait_for_clear_finished in the log, we'll look at the timestamp, add `timedelta(seconds=0)`, we'll map it to `["complete"]["start"]` So each line in the events log we json serialize and check for mappings.

Then there's a big block of code to apply the offsets, but only if it doesn't make end < start. It also encodes the timestamps in a way we can use to insert into the links.

Then for each of the first keys in the output dict (in this case, just "complete" and "stable"), we add in the links (grafana and vmui).

Finally, I guess we don't return the dict, but we could. Instead we just log it to the events.log. We can look in events.log and find the links. You can click on the links if your terminal supports it, or as I did you can make a shell script to recursively search a folder and open the links.

AlbertoSoutullo (approved):
Ok, thanks for the detailed explanation. I am getting more familiar with this part of the code.

I think it is a good moment to start going the tasks I proposed in #124 

I will do 1 and 2, because I created some of those tools, and you could do 3. Will create the tasks in the roadmap.

PR #130, opened 2025-08-13
Replace requirements.txt with uv and cleanup

AlbertoSoutullo:
As mentioned, we had 2 different `requirements.txt` for different modules, while also having code that was not used, with problematic dependencies (jupyter notebooks).

I removed code that was not used (visualizer, mesh creator, old scripts), and replaced the requirements.txt with uv.

This is faster, more consistent thanks to the lock file, and it automatically installs the required python and the dependencies with only one command.

On a side note, I think now also is the right time to create a single `main.py` that uses deployment and analysis, so we can have consistent imports. Right now if you want to use analysis or deployment, it is a bit dirty and manual to go to that directory, otherwise you have problems with the imports.


pyproject.toml
```diff
@@ -0,0 +1,18 @@
...
+    "dateparser>=1.2.2",
+    "httpx>=0.28.1",
+    "humanfriendly>=10.0",
+    "kubernetes>=33.1.0",
+    "pandas>=2.3.1",
+    "pydantic>=2.11.7",
+    "pyyaml>=6.0.2",
+    "requests>=2.32.4",
+    "result>=0.17.0",
+    "ruamel-yaml>=0.18.14",
```
PearsonWhite, line 16:
```suggestion
    "ruamel-yaml==0.17.21",
```
The removed `safe_load` in newer versions. I'll bump the version in a different PR.

pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[project]
+name = "10ksim"
+version = "0.1.0"
+description = "Add your description here"
```
PearsonWhite, line 4: Fill out description

PR #131, opened 2025-08-20
Make _metadata_event use dict events

PearsonWhite:
* Looks for events by comparing dicts

* Separate out parsing logic for metadata event

* Make class method _get_metadata_event


PearsonWhite:
This PR is motivated by a few things:
1. Mainly, it is meant to set up for another incoming PR where I can put the events in `BaseExperiment`.
2. Allows us to use more descriptive events by looking for dict key/value matches (eg. `{"event": "deployment", "phase": "start"}`)
3. It breaks up the logic into some helper functions that aren't instance methods. I tried to make the logic easier to understand, since it was pretty messy before. Also, this should help us extend/modify per-experiment.

deployments/deployment/waku/experiments/regression/regression.py
```diff
@@ -27,6 +32,80 @@
...
+                            path,
+                            new_value,
+                            sep=".",
+                        )
+                except KeyError:
+                    pass
+    return return_dict
+
+
+def format_metadata_timestamps(metadata) -> dict:
```
AlbertoSoutullo, line 73: Typing

deployments/deployment/waku/experiments/regression/regression.py
```diff
@@ -27,6 +32,80 @@
...
+    def format_item(node):
+        try:
+            return node.strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3] + "Z"
+        except AttributeError:
+            pass
+
+    return dict_apply(metadata, format_item)
+
+
+def get_valid_shifted_times(deltatime_map, metadata) -> dict:
```
AlbertoSoutullo, line 83: Typing

deployments/deployment/waku/experiments/regression/regression.py
PearsonWhite: Added

PR #134, opened 2025-08-21
Organize experiment output better

PearsonWhite:
* Remove --out-log flag

* Automatically pipe logging to out_folder/out.log

* Default out_folder to current datetime under ./out

* Copy values file into out_folder/cli_values.yaml


AlbertoSoutullo (approved):
Thanks!

PR #137, opened 2025-08-28
nimlibp2p mix node deployment changes

PearsonWhite:
Commits are on top of #131 and #132

### DIFF
--- a/deployments/deployment.py
+++ b/deployments/deployment.py
@@ -7,7 +7,7 @@ from kubernetes import config
 from kubernetes.client import ApiClient
 from ruamel import yaml
 
-from kube_utils import init_logger
+from kube_utils import init_logger, set_config_file
 from registry import registry as experiment_registry
 
 logger = logging.getLogger(__name__)
@@ -23,6 +23,7 @@ def run_experiment(
     if not kube_config:
         kube_config = "~/.kube/config"
     config.load_kube_config(config_file=kube_config)
+    set_config_file(kube_config)
     api_client = ApiClient()
 
     try:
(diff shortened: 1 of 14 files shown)


AlbertoSoutullo:
> `deployment/nimlibp2p/experiments/mix/regression.py` is mostly copied from `deployment/nimlibp2p/experiments/regression/regression.py`, but I decided that it was different enough to warrant having a different file. When reviewing, it may help to go to the commit before the formatting was done and compare the two files.

Thanks for the PR. It is kinda hard for me to follow why we need two different files for mix deployment. We are just talking about some more env variables, a volume (this should not matter) and then passing the wiper after the experiment (not dependant on mix deployment itself).

PearsonWhite:
> > `deployment/nimlibp2p/experiments/mix/regression.py` is mostly copied from `deployment/nimlibp2p/experiments/regression/regression.py`, but I decided that it was different enough to warrant having a different file. When reviewing, it may help to go to the commit before the formatting was done and compare the two files.
> 
> Thanks for the PR. It is kinda hard for me to follow why we need two different files for mix deployment. We are just talking about some more env variables, a volume (this should not matter) and then passing the wiper after the experiment (not dependant on mix deployment itself).

It could potentially be merged. The reasoning was that we're building two statefulsets, setting a bunch of env vars, and I haven't tested the logic with non-mixnet tests.

AlbertoSoutullo:
Thanks! When the deployment refactor is done, I think we definitely should merge it.

AlbertoSoutullo (approved):

PR #143, opened 2025-09-24
Build basic Waku nodes with python dict

PearsonWhite:
In order to break this task down into smaller parts, this is just the easiest way to change from building with Helm to a python dict. It just recreates the Helm template as a dict and changes nothing else. In the future, I expect:
- This will be broken into more readable pieces
- Probably wrapped in a WakuDeploymentBuilder class of some kind
- The config structure will be changed and made into a class (instead of dict/SimpleNamespace)
- The config merging logic will be removed to make application of the config to the template more straightforward

### DIFF
--- a/deployments/deployment/builders.py
+++ b/deployments/deployment/builders.py
@@ -7,9 +7,14 @@ from typing import List, Optional
 from ruamel.yaml import YAMLObject
 from ruamel.yaml.comments import CommentedMap
 
+from deployment.waku_nodes import waku_node
 from kube_utils import get_values_yamls, get_YAML, helm_build_dir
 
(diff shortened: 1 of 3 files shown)


PearsonWhite:
The scope of this issue was to determine whether or not this approach is feasible. I would say that it is.
It seems like a giant block of Helm template code just in Python dict form, but as I've added different things for other nodes (not in this branch), it gets broken up into more manageable functions. It starts to look more and more like:
```
deployment_yaml = {
     # Less giant block of code, because it only has all the common properties
}
init_containers = waku_init_containers(config)
if init_containers:
    out_deployment_dict["spec"]["template"]["spec"]["initContainers"] = init_containers
```

PR #164, opened 2025-10-17
Allow multiple match groups for the same query in victoria_reader.py

PearsonWhite:


PearsonWhite:
> Are you sure this doesn't break anything? What kind of scenario can we have that we have two match groups for the same query?

This is because in #163 there are two specific lines with "Failed.*?...", and then a generic "Failed" to catch any failure. Without allowing all tracers to be checked, any failure that matches the first won't show up on the generic failure.
https://github.com/vacp2p/10ksim/blob/8f177dd5fd49dd4651bc1cdd487d311704b2dcad/analysis/src/mesh_analysis/analyzers/custom_analyzer.py#L79

PR #173, opened 2025-11-19
Add builders/waku for basic regression nodes

PearsonWhite:


deployments/builders/libp2p/helpers.py
```diff
@@ -0,0 +1,40 @@
...
+    if ($2 == 1.0) {
+      exit 0;  # success, healthy state
+    } else {
+      exit 1;  # failure, unhealthy state
+    }
+  }
+  END { if (NR == 0) exit 1 }  # If no matching line is found, exit with failure
+'
+"""
+    return prefix + [command_block]
```
AlbertoSoutullo: We need to remember that this only works when there is only one topic. I think it would be good either modify the string with the nº of topis in `if ($2 == 1.0)` or at least leave a comment here as reminder.

AlbertoSoutullo (changes requested):
Minor comments, but looks very good.

deployments/builders/libp2p/helpers.py
PearsonWhite: Changed in 4e6799c4605735e5d6a970d27ff51ba3eb2822a4

deployments/builders/libp2p/helpers.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def readiness_probe_metrics():
+    return V1Probe(
+        _exec=V1ExecAction(command=readiness_probe_command_metrics()),
```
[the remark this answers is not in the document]
PearsonWhite: Changed in 4e6799c4605735e5d6a970d27ff51ba3eb2822a4

deployments/builders/waku/nodes.py
```diff
@@ -0,0 +1,136 @@
...
+        return V1ResourceRequirements(
+            requests={
+                "memory": "64Mi",
+                "cpu": "150m",
+            },
+            limits={
+                "memory": "600Mi",
+                "cpu": "400m",
+            },
+        )
```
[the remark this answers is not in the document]
PearsonWhite: Also same as the https://github.com/vacp2p/10ksim/pull/173#discussion_r2546626889 above.

deployments/builders/waku/builders.py
```diff
@@ -0,0 +1,126 @@
...
+    def __init__(self, config: StatefulSetConfig):
+        super().__init__(config)
+
+    def build(self):
+        if not self.config.name:
+            raise ValueError(f"Must configure node first. Config: `{self.config}`")
+        return super().build()
+
+    def with_regression(self) -> Self:
+        self.with_nice_command(19)
```
AlbertoSoutullo: We can leave `with_nice_command` but I think we shouldn't use it. I didn't question this when this was included in the past, but just looked at it and it means "runs the command with a very low CPU priority." which I don't think it makes sense. Should be the oposite actually. I would comment this line and stop using nice from now on...

PearsonWhite: Removed `with_nice_command` from `with_regression` and added a docstring. (982a246e121e458e75d20dee38037f530fc3c26e)

deployments/builders/waku/enr_or_addr.py
```diff
@@ -0,0 +1,130 @@
...
+        "imagePullPolicy": "IfNotPresent",
+        "volumeMounts": [
+            {
+                "name": f"{type_}-data",
+                "mountPath": f"/etc/{defaults['shortname']}",
+            }
+        ],
+        "command": [f"/app/get{type_}.sh"],
+        "args": [str(num)] + service_names,
+    }
```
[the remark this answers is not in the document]
PearsonWhite: If you wanted to add a new initContainer, it would probably be easiest to:
1. Create it as a dict.
2. Add it with `config.add_init_container`.
3. Which would be done in the `Waku/builder` with the `add_init_container` from `PodSpecBuilder` (added in 982a246e121e458e75d20dee38037f530fc3c26e)

I think we'll end up changing the addr/enr init_container later, since I already had to make a custom one for jswaku.

PR #187, opened 2026-01-14
Reorganize repo by feature

PearsonWhite:
- Change builders/waku -> waku/builders, etc

- Put utils and base builders in core/

- Fix imports


ufarooqstatus (approved):
Looks good to me!

PR #199, opened 2026-02-08
Format and add formatting hook

PearsonWhite:
* Add check_format script

* Add hooks/pre-push

* Add install for git hook in Makefile

### DIFF
--- a/deployments/deployment/waku/experiments/jswaku/jswaku.py
+++ b/deployments/deployment/waku/experiments/jswaku/jswaku.py
@@ -20,7 +20,6 @@ from experiments.base_experiment import (
 )
 from kubernetes import client
 from kubernetes.client import ApiClient
-from pod_api_requester.pod_api_requester import publish_message
 from pydantic import BaseModel, ConfigDict
 from registry import experiment
 from ruamel import yaml


deployments/Makefile
```diff
@@ -1,4 +1,11 @@
+.PHONY: format check install-hooks
+
+install-hooks: mkdir -p .git/hooks cp hooks/pre-push .git/hooks/pre-push chmod +x .git/hooks/pre-push
+
 format:
 	autoflake --in-place --remove-all-unused-imports --recursive .
 	isort --profile black -l 100 .
-	black -l 100 .
\ No newline at end of file
+	black -l 100 .
```
AlbertoSoutullo: This `-l` could be added to pyproject.toml so it is also visible for the IDEs.

deployments/Makefile
```diff
@@ -1,4 +1,11 @@
+.PHONY: format check install-hooks
+
+install-hooks: mkdir -p .git/hooks cp hooks/pre-push .git/hooks/pre-push chmod +x .git/hooks/pre-push
```
AlbertoSoutullo: `.git` is in an upper directory. Or actually, `check_format.sh` and Makefile should go in the parent directory.

Makefile
```diff
@@ -1,4 +1,11 @@
+.PHONY: format check install-hooks
+
+install-hooks: mkdir -p .git/hooks cp hooks/pre-push .git/hooks/pre-push chmod +x .git/hooks/pre-push
+
 format:
 	autoflake --in-place --remove-all-unused-imports --recursive .
 	isort --profile black -l 100 .
-	black -l 100 .
\ No newline at end of file
+	black -l 100 .
```
PearsonWhite: Added in d63b1dd01984a7cce35f29e19dd510002a37ec23

deployments/Makefile
PearsonWhite: Moved in 6c9966d7a8894b9328cdf241fbb0ec50b678ca67 and a7d2aa5605ffa5f988b232bc4ae0b1c5ec6746b2
Then I re-ran the formatter from the root dir (30290cf9fb56ba577c946811954febcd69952db3)

PR #201, opened 2026-02-08
Add annotations for PodConfig

PearsonWhite:
Annotation for config builders from #192


AlbertoSoutullo (commented):
Is this ready for review? Also, what was the context for the annotation?

PR #216, opened 2026-03-26
Analysis refactor

### LINKED ISSUE #202: Analysis yaml refactor
[Roadmap Link](https://roadmap.vac.dev/dst/ift/2026q1-dst-tooling#analysis-yaml-refactor)

#### Description
Refactor the analysis tools to get rid of the configuration yaml files, and use a similar pattern
to establish with analysis needs to be done after the deployment.

PearsonWhite:
# Tracers

## Changes
```
MessageTracer:
  Now, instead of:
    _patterns: List
    _tracings: List
    _extra_fields : List
  We have:
    patterns: List[PatternGroup] = Field(default_factory=list)
```
It's easier to see the changes with an example, so I've added one below.

## Before:
```
MessageTracer:
    _patterns:
      r'received relay message...',
      r'handling lightpush request.*?topics="waku lightpush legacy"...'
      r'handling lightpush request.*?my_peer_id=([\w*]+)...'
    _tracings:
      self._trace_received_in_logs
      self._trace_legacy_lightpush_in_logs
      self._trace_lightpush_in_logs

# Then VictoriaReader would need to be built with a query that matches what the tracer has.
# For that to happen, VictoriaReaderBuilder needs be built with a query that matches what the tracer has.
# So inside our WakuAnalyzer, we had:
class WakuAnalyzer:
  def _analyze_reliability_cluster(self, n_jobs: int):
      # Query gets passed to reader builder, then reader, then should match up with tracer patterns and tracings.
      queries = ["(received relay message OR  handling lightpush request)", "sent relay message"]
      reader_builder = VictoriaReaderBuilder(tracer, queries, **self._kwargs)
```

## After

Now, when we add to our tracers, it includes everything:
```
class WakuTracer:
  def with_received_pattern_group(self):
    # It adds all components: query, regexes (patterns) and conversion functions (tracings)

  patterns: List[PatternGroup]
    - PatternGroup:
      name: "received"
      query: "(received relay message OR  handling lightpush request)"
      trace_pairs:
        - regex: r'received relay message...'
          convert: self._trace_received_in_logs
        - regex: r'handling lightpush request.*?topics="waku lightpush legacy"...'
          convert: self._trace_legacy_lightpush_in_logs
        - regex: r'handling lightpush request.*?my_peer_id=([\w*]+)...'
          convert: self._trace_lightpush_in_logs,
```
It's true that the tracer.query would potentially be useless for some other pull method (for example, reading from a local file). It can be ignored in such cases. Otherwise, for cases where it is nessesary, like pulling from VictoriaLogs, it makes the code less error prone.

## MessageTracer.trace

Now, MessageTracer returns `dict` instead of `list` for trace results.
Note: We switch between WakuAnalyzer and Nimlibp2pAnalyzer here. Just ignore that.
```
# Before
MessageTracer:
  def trace(self, parsed_logs: List[List]) -> List[List]:
    # Returns a list of lists relying on the order matching for all:
    #  - parsed_logs
    #  - self._patterns
    #  - self._tracings

class WakuAnalyzer:
    def _analyze_reliability_cluster(self, n_jobs: int):
      dfs = self._merge_dfs(dfs)

  def merge_dfs(...):
    # For each pod we've scraped data from,
    # get the first in the list.
    received_df = pd.concat(
      [pd.concat(group[0], ignore_index=True) for group in dfs], ignore_index=True
    )

# After
class MessageTracer:
    def trace(self, parsed_logs: List[List[Tuple]]) -> Dict[str, List[pd.DataFrame]]:
      # Returns dict of list based on the structure in the tracer.

class Nimlibp2pAnalyzer(Analyzer):
    def _analyze_reliability_cluster
      dfs = self._merge_dfs(dfs, has_shards)

  def _merge_dfs(...):
    # For each pod we've scraped data from,
    # get data corresponding to the name "received".
    received_df = pd.concat(
        [pd.concat(group["received"], ignore_index=True) for group in dfs],
        ignore_index=True,
    )
  # For each pod we've scraped data from, get the first in the list.
```
Again, this prevents having to keep track of the order for queries, regex, etc. across multiple layers of code.
I did keep the `MessageTracer.trace()` the same in a lot of ways, so we still have the same `get_dataframes` in `VictoriaReader`
```
    def get_dataframes(self) -> Dict[str, List[pd.DataFrame]]:
        results = self.make_queries()
        dfs = self._tracer.trace(results)
```

# VictoriaReaderBuilder
## Before
We passed `queries` to VictoriaReaderBuilder and built the queries in each function based on what we passed in.
For example:
```
  def build_with_queries(
      self,
      stateful_set_name: str,
      node_index: Optional[int] = None,
      uniq_by: Optional[str] = None,
  ) -> VictoriaReader:
    # Builds query with stateful_set_name and node_index.
```
## After
Our new VictoriaReaderBuilder is quite similar, except:
1. The logic for query construction is relegated to `VictoriaQueryBuilder`.
   This gives us a consistent way to add params to the query throughout all functions.
2. TODO: Instead of passing `queries` manually, it can be already encapsulated in the tracer.
        This is not done yet. This can be done in some cleanup for DataPuller and the Reader/Builders.
        `src/analysis/mesh_analysis/analyzers/data_puller.py:54`
```
  def build_with_queries(
      self,
      stateful_set_name: str,
      node_index: Optional[int] = None,
      uniq_by: Optional[str] = None,
  ) -> VictoriaReader:
    query_builder = self._query_builder(uniq_by).with_stateful_set(stateful_set_name, node_index)
```

# Analyzers

## Analyzer structure
```
class Analyzer(BaseModel):
    _analysis_steps: List[AnalysisStep] = []
    data_puller: Optional[DataPuller] = None
    dump_analysis_dir: Optional[str] = None
```

Analyzers now have a base class that defines the following:
## AnalysisStep
AnalysisStep: This is basically just a big wrapper around a lambda function. This way we can:
1. Define what to do on failure
2. Name/organize each "test" (AnalysisStep action)
3. Give a consistent class for the result of a "test"

This helps break up long analysis functions into smaller pieces. One example of using this:
### Before
`WakuAnalyzer._analyze_reliability_cluster` calls `WakuAnalyzer._assert_num_nodes` and then goes on to do analysis.
WakuAnalyzer._assert_num_nodes reference: `src/analysis/mesh_analysis/analyzers/waku/waku_analyzer.py:159`
### After
`Nimlibp2pAnalyzer.check_ss_nodes` is its own AnalysisStep.
Nimlibp2pAnalyzer.check_ss_nodes reference: `src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py:84`

### "Builders"
This means instead of directly calling the each "test", we add the AnalysisSteps, and running them is managed in `log_multi_analysis.py`. So at the top of our Analysis classes we have `with_` functions (eg. `with_reliability_check`). This is how we define what tests an Analyzer will run.

## AnalysisResult
AnalysisResult: Similar to the above. Provides a structure for printing results. Contains:
- name
- status
- intermediates: dict so we can pass up information from inside the test, such as error results or number of nodes in a StatefulSet (like we see in `check_ss_nodes`).
```
class AnalysisResult(BaseModel):
    name: str
    intermediates: dict
    status: Literal["passed", "failed", "skipped", "error"]
```

## data_puller
Instead of creating the objects to pull data in the class, we inject that into each individual analyzer.

### Before

```
class WakuAnalyzer:
  def _analyze_reliability_cluster(self, n_jobs: int):
    tracer = ...
    queries = ...
    reader_builder = VictoriaReaderBuilder(tracer, queries, **self._kwargs)
    stack_analysis = VaclabStackAnalysis(reader_builder, **self._kwargs)
    dfs = stack_analysis.get_all_node_dataframes(n_jobs)
    # ... do analysis
```

### After

```
class Nimlibp2pAnalyzer(Analyzer):
  def analyze_reliability
    tracer = self.reliability_tracer(extra_fields)
    reliability_result = self._analyze_reliability_cluster(...)

  def _analyze_reliability_cluster(...) -> MessageReliabilityResult:
    # self.data_puller was already passed in,
    # We no longer create the VictoriaReaderBuilder and VaclabStackAnalysis in each analysis function.
    dfs = self.data_puller.get_all_node_dataframes_new(tracer, stateful_sets, nodes_per_ss)
    # ... do analysis
```

We also pass in the expected nodes from outside the analyzer for the ss_check.


src/analysis/mesh_analysis/analyzers/analyzer.py
```diff
@@ -0,0 +1,117 @@
...
+    _analysis_steps: List[AnalysisStep] = []
+    data_puller: Optional[DataPuller] = None
+    dump_analysis_dir: Optional[str] = None
+
+    def run(self) -> List[AnalysisResult]:
+        self._set_up_paths()
+        results = []
+        for analysis in self._analysis_steps:
+            result = analysis.run()
+            if result.status in ["error", "fail"] and analysis.on_fail != "continue":
```
AlbertoSoutullo: You are using `failed` instead of `fail`. Change the one you consider.

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
```diff
@@ -25,180 +44,180 @@ class Nimlibp2pAnalyzer:
...
+            dfs[1],
+            self._dump_analysis_path,
         )
-        received_df.set_index(["latency"], inplace=True)
-        received_df.sort_index(inplace=True)
 
-        return received_df
+        if reliability_results.nodes_missing_messages:
+            logger.info("Dumping logs from nodes with issues")
+            self._dump_logs(reliability_results.nodes_missing_messages)
```
AlbertoSoutullo, line 200: I think this method doesn't exist now. Will it be added in an upcoming PR?

src/analysis/mesh_analysis/analyzers/waku/waku_analyzer.py
```diff
@@ -367,6 +62,9 @@ def check_store_messages(self):
         :return:
         """
         waku_tracer = WakuTracer().with_wildcard_pattern()
+
+        self.stack.get_pod_logs(pod_identifier="get-store-messages", query="*")
```
AlbertoSoutullo: `self.stack` is not defined anywhere. Upcoming PR?

src/analysis/mesh_analysis/analyzers/data_puller.py
```diff
@@ -0,0 +1,112 @@
...
+        elif self._source_type == "local":
+            reader = FileReader(self._local_folder, tracer, self._jobs)
+            dfs = reader.get_dataframes()
+            return dfs
+        raise NotImplementedError()
+
+    def get_all_node_dataframes(self, tracer: MessageTracer, queries: Optional[str] = None):
+        if self._source_type == "victoria":
+            stack = self._make_stack(tracer, queries)
+            return stack.get_all_node_dataframes(self._jobs)
```
AlbertoSoutullo: There are missing parameters here right?

src/deployments/experiments/base_experiment.py
```diff
@@ -258,6 +258,16 @@ def _dump_metadata(self):
...
+    def log_metadata(self, metadata: dict):
+        self.log_event({**{"event": "metadata"}, **metadata})
+
+    def _dump_metadata(self):
+        metadata = self._get_metadata()
+        self.log_metadata(metadata)
+        out_path = Path(self.metadata_log_path)
+        with open(out_path, "a") as out_file:
+            out_file.write(json.dumps(metadata, default=str))
+
```
AlbertoSoutullo: This is duplicated.

AlbertoSoutullo (changes requested):
Left some comments. I am still not sure why we want WakuAnalyzer to inherit from Nimlibp2pAnalyzer.

src/deployments/experiments/base_experiment.py
PearsonWhite: Cleaned up in 92d30aa90b71fd20cc37525ad9592b3855fe73b0

src/analysis/mesh_analysis/analyzers/data_puller.py
PearsonWhite: 92d30aa90b71fd20cc37525ad9592b3855fe73b0
This was replaced by `get_all_node_dataframes_new`.
Removed `get_all_node_dataframes`
Renamed `get_all_node_dataframes_new` -> `get_all_node_dataframes`

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
PearsonWhite: Added this function in 92d30aa90b71fd20cc37525ad9592b3855fe73b0

src/analysis/mesh_analysis/analyzers/analyzer.py
PearsonWhite: Fixed in 92d30aa90b71fd20cc37525ad9592b3855fe73b0

AlbertoSoutullo (approved):
Thanks!

src/analysis/mesh_analysis/analyzers/waku/waku_analyzer.py
PearsonWhite: Fixed in 4206afe61cf8181619f36d5585a6ca1013f318e5

PR #239, opened 2026-04-09
Scrape refactor

### LINKED ISSUE #202: Analysis yaml refactor

PearsonWhite:


scrape.py
```diff
@@ -1,21 +1,144 @@
...
+
+    filters = []
+    if experiment_class:
+        filters.append(filter_by_class)
+
+    paths = [folder / path for path in get_folders(Path(folder), "metadata.json")]
+    for d in paths:
+        logger.info(d)
+        for exp in extract_exps(paths, filters):
+            yield exp
```
AlbertoSoutullo: Hm this feels weird. What `d` stands for?
We are filtering just by `NimLibp2pExperiment` class but what if inside we don't want to plot everything?
I don't quite follow this list comprehension + double for loop.

src/analysis/metrics/libp2p/metrics.py
AlbertoSoutullo, line 1: Why in some places we have TODO namespace? Can't we add it now?

src/analysis/plotting/config.py
```diff
@@ -0,0 +1,74 @@
...
+            if isinstance(folder, Path):
+                folder = folder.as_posix()
+            if folder.endswith("/"):
+                return folder
+            else:
+                return f"{folder}/"
+
+        folders = [ensure_trailing_slash(folder) for folder in folders]
+        self.config.folder.extend(folders)
+        return self
```
AlbertoSoutullo, line 62: This list comprehension `[ensure_trailing_slash(folder) for folder in folders]` doesn't make sense if we are accepting folders as `str` or `Path`.

src/analysis/metrics/scrape_utils.py
```diff
@@ -28,6 +30,7 @@ def create_promql(
 def get_query_data(request: str) -> Result[Dict, str]:
     response = urllib.request.urlopen(request, timeout=30)
 
+    print(f"Response: {response.status}")
```
AlbertoSoutullo: Use logger in debug likely

src/analysis/metrics/libp2p/metrics.py
```diff
@@ -0,0 +1,109 @@
+from typing import Iterator
+
+from src.analysis.metrics.config import MetricToScrape
+
+
+class Libp2pMetrics:
```
AlbertoSoutullo: A class with only staticmethods is a bit smelly to me.
Why don't we add namespace as init parameter to the class, everything stops bein static, and we can reuse namespace everywhere?

AlbertoSoutullo (changes requested):

src/analysis/metrics/libp2p/metrics.py
PearsonWhite: Ah yes. This used to be in a different file when I was writing this, so I used a class to contain it.

Now, I think we should either:
1. Remove the `Libp2pMetrics` class and just have functions.
2. Have instance methods.

I'm actually leaning towards 1, since the object doesn't serve much of a purpose. What do you think?

PearsonWhite: I did 1 in a9b45d3ee56548065b9907868f101ab15c790ee1

src/analysis/plotting/config.py
PearsonWhite: Fixed in cfe8009bcba7e4d09493eb15e3b2def8d567e087

src/analysis/metrics/libp2p/metrics.py
PearsonWhite: Fixed in cfe8009bcba7e4d09493eb15e3b2def8d567e087

src/analysis/metrics/scrape_utils.py
PearsonWhite: Fixed in cfe8009bcba7e4d09493eb15e3b2def8d567e087

scrape.py
PearsonWhite: Good catch. This was supposed to be this:
```
for d in paths:
    logger.info(d)
for exp in extract_exps(paths, filters):
    yield exp
```
And the `d` was for "debug".

Removed in cfe8009bcba7e4d09493eb15e3b2def8d567e087 since we print the experiment paths later anyways.

AlbertoSoutullo (approved):
Thanks!

PR #244, opened 2026-04-21
Nim-Libp2p KAD DHT Test 

### LINKED ISSUE #240: Closeness score
[Roadmap Link](https://roadmap.vac.dev/dst/logos/2026q2-anoncomms-evaluation#closeness-score)

#### Description
Design a way to obtain closeness score from KAD-DHT experiments.
If needed, the Kad-dht script in [libp2p-test-node](https://github.com/vacp2p/dst-libp2p-test-node)
repository can be modified, but shouldn't be necessary. PeerID is already exposed in the script already, so we need:
1. Deploy the experiment.
2. After the experiment is finished, scrape the logs of all PODs, searching for every PeerID in the network.
3. In the scripts for the KAD-DHT analysis, add this score as part of the analysis ifself.
4. Provide a way of visually represent this score.
For more information, refer to [this document](https://www.notion.so/KAD-DHT-planning-3098f96fb65c801eb80ccc0b3f395eb6?source=copy_link#3318f96fb65c80d8bf26fe0316691b0a).

radiken:
This PR facilitates nim-libp2p KAD DHT performance tests on kubernetes. Documentations for the test can be found here:
- [Test plan](https://www.notion.so/KAD-DHT-Planning-3098f96fb65c801eb80ccc0b3f395eb6)
- [Test results](https://www.notion.so/KAD-DHT-Result-3498f96fb65c80d993d8d2cff989c141)

The major changes include:
- yamls and script for depolying the test
- analysis scripts and helpers


AlbertoSoutullo (commented):
I general I think we should set up this as an experiment with the experiment decorator as the framework is inteded. I mean this is useful for quick testing but if we want to merge it would be nice to do it in the propper way. Could you help with this @PearsonWhite ?

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/nodes.yaml
```diff
@@ -0,0 +1,57 @@
...
+  replicas: 160
+  podManagementPolicy: "Parallel"
+  selector:
+    matchLabels:
+      app: zerotenkay
+  template:
+    metadata:
+      labels:
+        app: zerotenkay
+        role: nodes
```
PearsonWhite, line 16: How is the "role: nodes" label used? Does the test-node filter on that somehow?

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/bootstrap.yaml
```diff
@@ -0,0 +1,71 @@
...
+              cpu: "400m"
+          env:
+            - name: NODE_ROLE
+              value: "RoleBootstrap"
+            - name: PORT
+              value: "5000"
+            - name: DISCOVERY
+              value: "kad-dht"
+---
+apiVersion: v1
```
PearsonWhite, line 58: It think it would be easier to have this in a separate yaml.

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
```diff
@@ -427,3 +429,71 @@ def _check_if_msg_has_been_sent(
                 )
 
         return messages_sent_to_peer
+
```
PearsonWhite: I think we should have this in another class. Then we could use it as a mixin if we want it in the same Analyzer. I don't see anywhere that we're using any Nimlibp2pAnalyzer functions, so it should be safe to do this.

src/analysis/mesh_analysis/readers/tracers/nimlibp2p_tracer.py
```diff
@@ -94,3 +94,117 @@ def _create_dataframe_with_timestamp(self, parsed_logs: List[str], columns: List
             ) from e
 
         return df
+
+    def with_kad_dht_pattern(self) -> Self:
```
PearsonWhite: This too we can have in its own tracer.

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
```diff
@@ -427,3 +429,71 @@ def _check_if_msg_has_been_sent(
...
+    def get_bootstrap_start_time(self, bootstrap_pod: str = "bootstrap-0") -> pd.Timestamp:
+        tracer = Nimlibp2pTracer(extra_fields=self._kwargs.get("extra_fields")).with_node_started_pattern()
```
PearsonWhite: `self._kwargs` doesn't exist anymore. If your analysis relies on extra fields those will be here, with the rest of the analysis logic. See `src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py:145`

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
```diff
@@ -427,3 +429,71 @@ def _check_if_msg_has_been_sent(
...
+        queries = ['"Node started"']
```
PearsonWhite:
```suggestion
```
The queries are not embedded in the tracer, so you don't need this.

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
```diff
@@ -427,3 +429,71 @@ def _check_if_msg_has_been_sent(
...
+        reader_builder = VictoriaReaderBuilder(tracer=tracer, queries=queries, kwargs=self._kwargs)
+        stack_analysis = VaclabStackAnalysis(reader_builder, **self._kwargs)
```
PearsonWhite:
```suggestion
        df = self.data_puller.get_pod_logs(tracer, bootstrap_pod, tracer.patterns[0].query)[0]
```
I think this should suffice here. I haven't tested though. I see some things that need some cleaning in DataPuller, so just use whatever works here. `VictoriaReaderBuilder` and `VaclabStackAnalysis` are fine for now, just make sure they're imported.

PearsonWhite: So I think the whole thing would look something like this:
```
        extra_fields = ["kubernetes.pod_name", "kubernetes.pod_node_name"]
        tracer = Nimlibp2pTracer(extra_fields=extra_fields).with_node_started_pattern()
        reader_builder = VictoriaReaderBuilder(tracer=tracer, queries=tracer.patterns[0].query, extra_fields=extra_fields, kwargs=self.data_puller.kwargs)
        stack_analysis = VaclabStackAnalysis(reader_builder, self.data_puller.kwargs)
```
Since the DataPuller would pass the kwargs to the VaclabStackAnalysis for the url, etc., you would have to store the kwargs from `kad.py` in `Nimlibp2pAnalyzer.data_puller` to propagate them.

src/analysis/mesh_analysis/analyzers/analyzer.py
```diff
@@ -60,6 +60,7 @@ def run(self) -> AnalysisResult:
 
 class Analyzer(BaseModel):
     _analysis_steps: List[AnalysisStep] = []
+    _kwargs: dict = {}
```
PearsonWhite:
```suggestion
```
Instead of passing through here, pass it down through the data_puller.kwargs.

src/analysis/kad.py
```diff
@@ -0,0 +1,296 @@
...
+        "nodes_per_statefulset": [120, 1, 1],
+        "container_name": "node",
+        "namespace": "nimlibp2p",
+        "extra_fields": ["kubernetes.pod_name", "kubernetes.pod_node_name"],
+    }
+
+    log_analyzer = Nimlibp2pAnalyzer(
+        dump_analysis_dir="local_data/simulations_data/kad-dht/",
+    ).with_kwargs(stack)
+    log_lines = log_analyzer.check_kad_dht_result()
```
PearsonWhite: Why are we returning the log lines here and analyzing them outside the analyser?

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
PearsonWhite: That would also be a good place to document what we're analysing.

src/analysis/warmup.py
```diff
@@ -0,0 +1,122 @@
...
+
+sns.set_theme()
+
+
+def extract_node_index(pod_name: str) -> int:
+    """Extract numeric index from a pod name like 'nodes-47'."""
+    return int(pod_name.split("-")[-1])
+
+
+if __name__ == "__main__":
```
PearsonWhite: Why do we have a separate entrypoint? Not saying it's necessarily wrong btw.

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/deployment.sh
```diff
@@ -0,0 +1,14 @@
...
+kubectl rollout status --watch --timeout=600s statefulset/bootstrap -n $NAMESPACE
+
+kubectl apply -f nodes.yaml
+kubectl rollout status --watch --timeout=600s statefulset/nodes -n $NAMESPACE
+
+kubectl apply -f probe.yaml
+kubectl rollout status --watch --timeout=600s statefulset/nodes -n $NAMESPACE
+
+sleep 60
+./cleanup.sh
```
PearsonWhite, line 12:
```suggestion
#!/bin/bash
NAMESPACE="nimlibp2p"
KUBECONFIG=<YOURKUBECONFIGPATH>

kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE apply -f service.yaml

kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE apply -f bootstrap.yaml
kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE rollout status --watch --timeout=600s statefulset/bootstrap -n $NAMESPACE

kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE apply -f nodes.yaml
kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE rollout status --watch --timeout=600s statefulset/nodes -n $NAMESPACE

kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE apply -f probe.yaml
kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE rollout status --watch --timeout=600s statefulset/nodes -n $NAMESPACE

sleep 60
./cleanup.sh
```

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/cleanup.sh
```diff
@@ -0,0 +1,3 @@
+kubectl delete -f bootstrap.yaml
+kubectl delete -f nodes.yaml
+kubectl delete -f probe.yaml
```
PearsonWhite:
```suggestion
NAMESPACE="nimlibp2p"
KUBECONFIG=<YOURKUBECONFIGPATH>

kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE delete -f bootstrap.yaml
kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE delete -f nodes.yaml
kubectl --kubeconfig $KUBECONFIG --namespace $NAMESPACE delete -f probe.yaml
```

PearsonWhite (changes requested):
I haven't gotten through everything yet, but I wanted to put up the comments I had so far.

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/bootstrap.yaml
AlbertoSoutullo: Gotta say despite this is irrelevant once build the experiment in python, I usually ended up following this approach at the end, because each statefulset/deployment you are doing, it goes with its correspondant service in the same file, and it is very handy.

src/analysis/kad.py
AlbertoSoutullo: He is using my first quick example for the analysis. The idea is now to help him integrate it as it should.

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/deployment.sh
radiken: Since we are migrating the experiment to the new framework, I guess we can remove this file completely?

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/cleanup.sh
radiken: Same as `deployment.sh`: since we are migrating the experiment to the new framework, I guess we can remove this file completely?

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/nodes.yaml
radiken: Test-node doesn't read the label, it relies on the NODE_ROLE env var. 
This role label is functional in the bootstrap node for routing traffic to the bootstrap node. (@AlbertoSoutullo am I correct?)
For the regular nodes and probe, this is just informational metadata to easily filter them in kubectl.

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/bootstrap.yaml
radiken: Yes this is now irrelevant

src/analysis/warmup.py
radiken: When building this analysis it was for a side problem and out of my own interest to get this data. But now I think we can indeed keep this analysis and I agree that we should merge it with `kad.py`

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
radiken: Thanks for the recommendation! Agree, moving all KAD DHT specific logic to a new KadDHTAnalyzer class

src/analysis/mesh_analysis/analyzers/analyzer.py
radiken: Understood, thanks

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
radiken: I understand the new structure now, thanks

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
radiken: I think you mean the queries are **now** embedded in the tracer? :)

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
radiken: I ended up cleaning up `DataPuller.get_pod_logs` so that it extracts the queries from the tracer. So I can completely remove `VictoriaReaderBuilder` and `VaclabStackAnalysis` from the analyzer and just rely cleanly on `self.data_puller.get_pod_logs()`

src/analysis/mesh_analysis/readers/tracers/nimlibp2p_tracer.py
radiken: Thanks, moving them

radiken:
@AlbertoSoutullo @PearsonWhite 
Apart from addressing the comments and applying the experiment and analysis to the new framework, I did some update on the framework itself to address some problems I met to run my experiment:

* **StatefulSetBuilder Crash:** Fixed a `TypeError` where `labels` and `selector_labels` were initialized as `None` instead of empty dictionaries. Corrected the nested assignment paths so labels are properly applied to both the `stateful_set_spec` and the internal `pod_template_spec_config`.
* **Resource Cleanup & Rollouts:** Updated `kube_utils.py` to correctly identify and delete `Service` resources during the cleanup phase, preventing namespace pollution. We also patched `wait_for_rollout` to gracefully handle `Service` kinds without crashing.

BTW, the line [here](https://github.com/vacp2p/10ksim/blob/master/src/deployments/libp2p/builders/helpers.py#L11), `LIBP2P_CONTAINER_NAME = "pod-0"` doesn't seem to be a good name to me, I didn't change it, but if you would like I would change it to something like `"node"`

src/analysis/mesh_analysis/analyzers/data_puller.py
```diff
@@ -88,10 +88,13 @@ def get_all_node_dataframes(
...
     def get_pod_logs(
-        self, tracer, pod_identifier: str, query: str, *, order_by: Optional[str] = None
+        self, tracer, pod_identifier: str, query: Optional[str] = None, *, order_by: Optional[str] = None
     ):
         if self._source_type == "victoria":
-            stack = self._make_stack(tracer, query)
+            if query is None:
+                stack = self._make_stack_new(tracer)
+            else:
+                stack = self._make_stack(tracer, query)
```
AlbertoSoutullo: What was the difference between `make_stack_new` and `make_stack`?

src/analysis/mesh_analysis/readers/tracers/nimlibp2p_tracer.py
```diff
@@ -94,3 +94,5 @@ def _create_dataframe_with_timestamp(self, parsed_logs: List[str], columns: List
             ) from e
 
         return df
+
+
```
AlbertoSoutullo: Delete this

AlbertoSoutullo (commented):
Minor comments. This looks much better now. Did reproduce the results with this?

src/analysis/mesh_analysis/analyzers/data_puller.py
radiken: `_make_stack_new` automatically extracts the queries from the tracer object, and `_make_stack` requires the queries to be explicitly passed in as an argument.

Also, `_make_stack_new` explicitly passes `extra_fields=self.kwargs["extra_fields"]` into the `VictoriaReaderBuilder`, `_make_stack` does not.

The intention behind `_make_stack_new` was to start relying on the tracer directly for the queries rather than forcing the caller to provide them.

Maybe is not the best code but I tried not to change the framework too much in this PR and use it as intended

radiken:
> Did reproduce the results with this?

Yes I did a run, it worked well

src/analysis/mesh_analysis/readers/tracers/nimlibp2p_tracer.py
radiken: Will see if Pearson has some other comments and push this along with those :)

src/analysis/kad.py
radiken: Thanks for pointing this out, this is also updated now

AlbertoSoutullo:
> > Did reproduce the results with this?
> 
> Yes I did a run, it worked well

Nice, GJ!. Lets wait for @PearsonWhite's review then.

src/analysis/mesh_analysis/analyzers/nimlibp2p_analyzer.py
PearsonWhite: I noticed some things that need to be cleaned up with the DataPuller while reviewing this. I think I made some change in there that might conflict, but that will be settled in a different PR. This is fine for this PR.

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/cleanup.sh
PearsonWhite: Theoretically, we don't need any of the `experiment_scripts`, but it can be helpful to have when debugging things because you already have the complete yaml without having to generate it. Since it's already here, I would just keep it in.

PearsonWhite:
> * **Resource Cleanup & Rollouts:** Updated `kube_utils.py` to correctly identify and delete `Service` resources during the cleanup phase, preventing namespace pollution. We also patched `wait_for_rollout` to gracefully handle `Service` kinds without crashing.

Keeping the services running was intentional. We don't need to redeploy services each time. They will persist between experiment runs. If we want to re-evaluate this, it should be in a different PR.


> BTW, the line [here](https://github.com/vacp2p/10ksim/blob/master/src/deployments/libp2p/builders/helpers.py#L11), `LIBP2P_CONTAINER_NAME = "pod-0"` doesn't seem to be a good name to me, I didn't change it, but if you would like I would change it to something like `"node"`

I think this was legacy, but let's leave it for now.

src/deployments/core/configs/service.py
```diff
@@ -0,0 +1,39 @@
+from typing import Dict, List, Optional
```
PearsonWhite, line 1: Let's extract out all the service changes and put them in their own PR. It's beyond the scope of this PR.

src/deployments/core/builders.py
```diff
@@ -66,6 +77,35 @@ def build(self) -> V1Pod:
         return build_pod(self.config)
 
 
+class ServiceBuilder(BaseModel):
```
PearsonWhite, line 80: Move changes to its own PR

src/deployments/core/kube_utils.py
```diff
@@ -326,7 +326,7 @@ def get_cleanup_resources(yamls: List[yaml.YAMLObject], types: Optional[List[str
         "Service": [],
     }
     types = (
-        types if types else ["Deployment", "StatefulSet", "DaemonSet", "ReplicaSet", "Pod", "Job"]
+        types if types else ["Deployment", "StatefulSet", "DaemonSet", "ReplicaSet", "Pod", "Job", "Service"]
```
PearsonWhite: Remove this change

src/deployments/core/kube_utils.py
```diff
@@ -599,6 +599,10 @@ def poll_rollout_status(
 
         return False
 
+    elif kind == "service":
```
PearsonWhite, line 604: Remove this change

src/deployments/experiments/base_experiment.py
```diff
@@ -288,7 +300,8 @@ async def run(
                     "args": vars(args),
                 }
             )
-            shutil.copy(args.values_path, os.path.join(workdir, "cli_values.yaml"))
```
PearsonWhite: We shouldn't need to change this here.

src/analysis/mesh_analysis/analyzers/data_puller.py
PearsonWhite: This is basically an oversight in the refactor task. I'll clean it up later.

PearsonWhite (commented):
Aside from the other comments:
1. Rebase onto origin/master. This should reduce the number of diffs compared to master.
2. `make format`

PearsonWhite:
> Keeping the services running was intentional. We don't need to redeploy services each time. They will persist between experiment runs. If we want to re-evaluate this, it should be in a different PR.

After talking with @AlbertoSoutullo , probably we do want the services to be part of the cleanup, but it should still be in another PR, so you can include that change when you make the new PR.

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/cleanup.sh
radiken: Sure, bringing it back

deployment-utilities/experiment_scripts/nimlibp2p/kad-dht/deployment.sh
radiken: keeping this file and using the suggested version

src/deployments/core/configs/service.py
radiken: No problem, changes related to this are all removed, will soon open another PR for that

PearsonWhite (approved):
Looks good. Approved.

AlbertoSoutullo (approved):

PR #250, opened 2026-04-27
Fix debug logging

PearsonWhite:
* Log settings are now passed to workers in ProcessPoolExecutor

* Use UTC for logging by default

* Added CLI options in log_multi_analysis.py and scrape.py

### DIFF
--- a/log_multi_analysis.py
+++ b/log_multi_analysis.py
@@ -1,3 +1,4 @@
+import argparse
 import asyncio
 import json
 import logging
@@ -5,7 +6,6 @@ import os
 import sys
 import traceback
 from collections import defaultdict
-from contextlib import contextmanager
 from pathlib import Path
 from typing import Any, Iterator, Optional, Self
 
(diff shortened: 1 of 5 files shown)


AlbertoSoutullo (commented):
`traffic.py` changes are also in a different PR right?
Also, why do we need this? `dump_node_logs` is for dumping logs from the node. Why are we setting our logging conf here?

PearsonWhite:
> `traffic.py` changes are also in a different PR right? Also, why do we need this? `dump_node_logs` is for dumping logs from the node. Why are we setting our logging conf here?

`dump_node_logs` still eventually makes VictoriaLogs queries. So if we want to debug what it's doing there, we have to apply the logging config to the tasks in the ProcessPoolExecutor. It will also let us use a consistent config (like UTC time) for the log messages during the log dump.

AlbertoSoutullo (approved):

PR #251, opened 2026-04-29
Add ServiceBuilder, readiness probes, and deployment utilities

radiken:
- Introduced `ServiceBuilder` and `ServiceConfig` to allow experiments to define and deploy Kubernetes Services.
- Updated `kube_utils.py` to gracefully handle `Service` resources during cleanup phases and rollout status polling.
- Added a `.with_readiness_probe()` method in `Libp2pStatefulSetBuilder` to allow Kubernetes to track the health and readiness of libp2p node containers.
- Added a `dump_yaml` helper method to `BaseExperiment`.


src/analysis/mesh_analysis/readers/victoria_reader.py
```diff
@@ -43,7 +43,7 @@ def _fetch_data(self, url: str, headers: Dict, params: Dict, extra_fields: List[
                         logger.info(line)
                         exit()
                     logs.append(
-                        (parsed_object["_msg"],) + tuple(parsed_object[k] for k in extra_fields)
+                        (parsed_object.get("_msg", ""),) + tuple(parsed_object.get(k, None) for k in extra_fields)
```
PearsonWhite: Why does this change need to be here? The "_msg" field should always exist, right? And same with any extra fields. So if not, it should raise an error, not silently ignore it?

AlbertoSoutullo (approved):
LGTM! Thanks c:

src/deployments/libp2p/builders/builders.py
```diff
@@ -53,3 +53,14 @@ def with_option(self, key, value) -> Self:
     def with_image(self, image: Image) -> Self:
         self.with_image_in_container(LIBP2P_CONTAINER_NAME, image, overwrite=True)
         return self
+
+    def with_readiness_probe(self, path: str, port: int, initial_delay_seconds: int = 2, period_seconds: int = 2) -> Self:
```
PearsonWhite: Is this actually used anywhere? I have some local changes in `src/deployments/libp2p/builders/nodes.py` to add `config.with_readiness_probe(readiness_probe_metrics())` in `create_container_config`. I assume we'd actually want to just pass in a readiness probe instead of each param of V1Probe.

radiken: It was used in the kad-dht experiment :) since I extracted the framework changes only so it is currently not used. 
But yes, I agree we can just pass in a readiness probe

src/analysis/mesh_analysis/readers/victoria_reader.py
radiken: I actually ran into a bug during a large DHT experiment where a single malformed log line from VictoriaMetrics was missing the `kubernetes.pod_name` field. Looks like VictoriaMetrics can occasionally return unstructured or incomplete log lines depending on the cluster's logging pipeline. 

The downstream regex parsers in the MessageTracer will naturally fail to match and skip these malformed lines anyway, so I think it is safer to pass None here.

Or we can do a try except block here to raise warning if `_msg` is missing?

src/deployments/experiments/base_experiment.py
```diff
@@ -245,6 +249,14 @@ def _get_out_path(self, path: Path, out_dir: Optional[str]) -> Path:
                 )
             return Path(out_dir) / path
 
+    def dump_yaml(self, obj, workdir: str, name: str):
+        out_path = Path(workdir) / name / f"{name}.yaml"
```
PearsonWhite:
```suggestion
        out_path = Path(workdir) / f"{name}.yaml"
```
While we're moving this anyways, we can fix this path.

src/analysis/mesh_analysis/readers/victoria_reader.py
PearsonWhite: I think the try/except with a warning is a better solution.

src/deployments/experiments/base_experiment.py
radiken: Updated, thanks

src/deployments/libp2p/builders/builders.py
radiken: fixed

src/analysis/mesh_analysis/readers/victoria_reader.py
radiken: added a try except block

PearsonWhite (approved):

PearsonWhite:
Approved! I would recommend always squash-merging, even though required on this repo. It just keeps history cleaner.

PR #256, opened 2026-05-04
Added WakuExperiment for regression tests

PearsonWhite:


src/deployments/experiments/waku.py
```diff
@@ -0,0 +1,185 @@
...
+    def log_event(self, event):
+        logger.info(event)
+        return super().log_event(event)
+
+    class ExpConfig(BaseModel):
+        model_config = ConfigDict(extra="ignore")
+
+        num_messages: NonNegativeInt = 20
+        num_nodes: NonNegativeInt = 20
+        num_boostrap_nodes: NonNegativeInt = 5
```
AlbertoSoutullo: Typo `num_bootstrap_nodes`

src/deployments/experiments/waku.py
```diff
@@ -0,0 +1,185 @@
...
+        api_client: ApiClient,
+        workdir: str,
+        args: Namespace,
+        values_yaml: Optional[dict],
+        stack: ExitStack,
+    ):
+        self.log_event("run_start")
+
+        config = self.ExpConfig(**values_yaml)
+        config.num_boostrap_nodes = 5
```
AlbertoSoutullo: This will not correctly overwrite if the yaml has the correct name.

src/deployments/experiments/waku.py
```diff
@@ -0,0 +1,185 @@
...
+                    name="waku-node",
+                    name_template=random_name,
+                    service="zerotesting-service",
+                    port=8645,
+                )
+                await waku_publish(namespace=namespace, target=target)
+            except PodApiApplicationError as e:
+                import pdb
+
+                pdb.set_trace()
```
AlbertoSoutullo: Please, clean this

src/deployments/experiments/waku.py
```diff
@@ -0,0 +1,185 @@
...
+                logger.error(f"PodApiApplicationError: {e} {traceback.format_exc()}")
+            except PodApiError as e:
+                import pdb
+
+                pdb.set_trace()
```
AlbertoSoutullo: Same

src/deployments/experiments/waku.py
```diff
@@ -0,0 +1,185 @@
...
+                logger.error(f"PodApiError: {e} {traceback.format_exc()}")
+            except Exception as e:
+                import pdb
+
+                pdb.set_trace()
```
AlbertoSoutullo: Same

src/deployments/experiments/waku.py
```diff
@@ -0,0 +1,185 @@
...
+                logger.error(f"Other exception: {e} {traceback.format_exc()}")
+
+            await asyncio.sleep(config.delay_after_publish)
+
+        self.log_event("publisher_messages_finished")
```
AlbertoSoutullo, line 169: I don't think its correct to log the event "finished" even when the publisher could fail.

AlbertoSoutullo (changes requested):

src/deployments/experiments/waku.py
PearsonWhite: Fixed in e888cf501959cb5f1a3e0abec8c434d9c643028d

src/deployments/experiments/waku.py
PearsonWhite: Fixed in e888cf501959cb5f1a3e0abec8c434d9c643028d

src/deployments/experiments/waku.py
PearsonWhite: Fixed in e888cf501959cb5f1a3e0abec8c434d9c643028d

src/deployments/experiments/waku.py
PearsonWhite: Fixed in e888cf501959cb5f1a3e0abec8c434d9c643028d

src/deployments/experiments/waku.py
PearsonWhite: Fixed in e888cf501959cb5f1a3e0abec8c434d9c643028d

src/deployments/experiments/waku.py
PearsonWhite: "publisher_messages_finished" is used to find the window for analysis (start/end times for Victoria). We would still probably want to be able to run the analysis even if a message failed to send.

We don't have much error handler in the deployments in general. Probably the best place to put that would be an analyzer that just scans the logs for "ERROR". This particular failure would be caught by the reliability analyzer.

AlbertoSoutullo (approved):

PR #257, opened 2026-05-04
Fix metadata bridge when namespace not used.

PearsonWhite:
Added elif so the else block wouldn't fail if no namespaces were used.

### DIFF
--- a/src/deployments/core/base_bridge.py
+++ b/src/deployments/core/base_bridge.py
@@ -256,6 +256,8 @@ class BaseBridge(BaseModel):
 
         if len(namespaces) == 1:
             metadata["namespace"] = next(iter(namespaces))
+        elif len(namespaces) == 0:
+            logger.info(f"No namespaces used.")
         else:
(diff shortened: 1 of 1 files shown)


AlbertoSoutullo (commented):
Hmm do we ever want to use multiple namespaces now that I think about?
I think this should be only one if and else. And we should abort if we are using several namespaces. Or am I missing something?

PearsonWhite:
> Hmm do we ever want to use multiple namespaces now that I think about? I think this should be only one if and else. And we should abort if we are using several namespaces. Or am I missing something?

We could probably imagine cases for it, at least for development testing, so it doesn't raise an error here. I think you would have to go out of your way to do it though.

As for this PR, it's a fix with building the metadata after the experiment has already run. In this particular case, if we have an experiment that doesn't deploy anything to Kubernetes (eg. MultiExperiment, which just has other experiments run but doesn't run anything itself).

AlbertoSoutullo (approved):

PR #258, opened 2026-05-04
Added MultiExperiment base class MultiNimlibp2p

PearsonWhite:


src/deployments/experiments/multi_experiment.py
```diff
@@ -0,0 +1,126 @@
...
+        BaseExperiment.add_args(subparser)
+
+    @classmethod
+    def add_args(cls, subparser) -> None:
+        subparser.add_argument(
+            "--name", type=str, required=False, help="Name of the experiment to run."
+        )
+        subparser.add_argument(
+            "--delay",
+            type=Optional[float],
```
AlbertoSoutullo: This is not a valid.

src/deployments/experiments/libp2p/multi_nimlibp2p.py
```diff
@@ -0,0 +1,81 @@
...
+            "connect_to": 10,
+        }
+
+    def get_params_list(self) -> List[dict]:
+        return [item for item in self.exp_params()]
+
+    def get_name_from_params(self, params: dict) -> str:
+        version = params["version"]
+        keys = ["num_nodes", "num_messages"]
+        used_keys = filter(lambda item: item in keys, params.items())
```
AlbertoSoutullo, line 78: Isn't this comparing (key,value) tuples against string keys so it always filters out everything?

AlbertoSoutullo (changes requested):

src/deployments/experiments/libp2p/multi_nimlibp2p.py
PearsonWhite: This is because we may pass it a set of params that doesn't actually include "num_nodes" or "num_messages". In that case, it won't try to use those for the name. This could be because the defaults are defined in `_run`.

src/deployments/experiments/multi_experiment.py
PearsonWhite: Fixed in 07a128024250d24009a76655100578e2076cdafd

AlbertoSoutullo (approved):

PR #259, opened 2026-05-04
add delay+jitter with initContainer logic

PearsonWhite:


src/deployments/core/configs/helpers.py
```diff
@@ -152,3 +155,18 @@ def convert_to_container_config(container: ContainerConfig | V1Container | dict)
...
+    delay: Union[str, NonNegativeInt],
+    jitter: Union[str, NonNegativeInt],
+):
+    return V1Container(
+        name="slowyourroll",
+        image="soutullostatus/tc-container:1",
+        image_pull_policy="IfNotPresent",
+        security_context=V1SecurityContext(capabilities=V1Capabilities(add=["NET_ADMIN"])),
+        command=[
+            f"tc qdisc add dev eth0 root netem delay {str(delay)}ms {str(jitter)}ms distribution normal",
```
AlbertoSoutullo, line 170: Hmmm I was thinking that maybe we could adapt this to make it more flexible but this can be done later. But something like maintaining the constant part ie `tc qdisc add dev eth0 root netem delay` and then the rest we could adapt it, but we can do this later maybe.

AlbertoSoutullo (approved):

src/deployments/core/configs/helpers.py
PearsonWhite: I'm not sure what you mean here.

AlbertoSoutullo: What I mean is that we have a hardcoded netem delay here, but there is no way to inject packet loss for example. We just can change the delay and hitter. But not important right now.

PR #260, opened 2026-05-04
Allow custom condition in wait_for_rollout

PearsonWhite:
* `pod_ready_condition` param allows custom logic in checking if a pod is ready

* `condition` param allows waiting for certain number or percentage of pods to be read for StatefulSet

This lets us wait for conditions like "75% of pods are ready", and more precisely define what it means for a pod to be ready.

Example of usage:
```
await self.deploy(api_client, stack, args, values_yaml, deployment=nodes, wait_for_ready=False)

def ready_condition(ready, total):
    return ready / total > 0.75

def pod_ready(pod: V1Pod):
    conditions = pod.status.conditions
    return all(
        any(condition.type == search_key and condition.status == search_value for condition in conditions)
        for search_key, search_value in [
            ("ContainersReady", "True"),
            ("Ready", "True"),
        ]
    )

await wait_for_rollout(nodes, api_client, timeout=3000, pod_ready_condition=pod_ready, condition=ready_condition)
```

### DIFF
--- a/src/deployments/core/kube_utils.py
+++ b/src/deployments/core/kube_utils.py
@@ -18,15 +18,16 @@ from datetime import datetime, timedelta
 from datetime import timezone as dt_timezone
 from functools import lru_cache, partial
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterator, List, Literal, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterator, List, Literal, Optional, Protocol, Tuple, Union
 
 import dateparser
 import ruamel.yaml
 from kubernetes import client, utils
-from kubernetes.client import ApiClient, V1Probe
+from kubernetes.client import ApiClient, ApiException, V1ObjectMeta, V1Pod, V1Probe
 from kubernetes.client.models import V1Node
 from kubernetes.client.rest import ApiException
 from kubernetes.utils import FailToCreateError
+from pydantic import NonNegativeInt
 from ruamel import yaml
 from ruamel.yaml.comments import CommentedMap, CommentedSeq
 
@@ -160,7 +161,7 @@ def _extract_kind_and_name(kube_yaml: dict) -> Tuple[str, str]:
 def _kind_api_map(operation: str) -> Dict[str, Tuple[str, str]]:
     kind_map = {
         "Deployment": ("apps", "deployment"),
-        "StatefulSet": ("apps", "statefulset"),
+        "StatefulSet": ("apps", "stateful_set"),
         "DaemonSet": ("apps", "daemonset"),
         "ReplicaSet": ("apps", "replicaset"),
         "Job": ("batch", "job"),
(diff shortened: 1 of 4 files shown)


src/deployments/core/kube_utils.py
```diff
@@ -530,76 +538,66 @@ def cleanup():
...
+            # By default, just check if pods are ready according to statefulset's available_replicas.
+            obj = client.AppsV1Api(api_client).read_namespaced_stateful_set(name, namespace)
+            desired = obj.spec.replicas or 0
+            available = getattr(obj.status, "available_replicas", None)
+            if available is None:
+                available = getattr(obj.status, "ready_replicas", 0)
+            return available, desired
+        pods = get_pods_for_statefulset(name, namespace, api_client)
+        available = [check_pod_condition(pod, pod_status_condition) for pod in pods]
+        return (sum(available), len(pods))
```
AlbertoSoutullo: Hm I am concerned about this.
Usually for large stateful sets, it takes some time even for scheduling the pods. If this is done too early, we could have a ready == total of 0==0, or any other weird scenario where currently created pods are ready but the controller is still creating more. How are we sure we will not face this issue?
Because I remember in bash this was automatically done with `kubectl rollout status --watch`. But here... you are creating your own rollout function it seems. Isn't there in the API this functionality?

src/deployments/core/kube_utils.py
```diff
@@ -663,6 +674,42 @@ async def wait_for_rollout(
...
+    selector = []
+    if sts.spec.selector and sts.spec.selector.match_labels:
+        selector.extend(f"{key}={value}" for key, value in sts.spec.selector.match_labels.items())
+
+    label_selector = ",".join(selector)
+
+    return core_api.list_namespaced_pod(
+        namespace=namespace,
+        label_selector=label_selector,
+    ).items
```
AlbertoSoutullo: This looks a bit unsafe because if label selector ends up being an empty list, we will retrieve all pods in the namespace.

AlbertoSoutullo (changes requested):

PearsonWhite:
Simplified custom conditions in 73bc65a3b81c1ce49aa5b871544e84b4a614a9ea

We can still do the same thing as before, but the internal code simply uses a single condition.
Example usage:
```
await self.deploy(
    api_client, stack, args, values_yaml, deployment=nodes, wait_for_ready=False
)

def pod_ready(pod: V1Pod) -> bool:
    conditions = pod.status.conditions
    if not conditions:
        return False
    return all(
        any(
            condition.type == search_key and condition.status == search_value
            for condition in conditions
        )
        for search_key, search_value in [
            ("ContainersReady", "True"),
            ("Ready", "True"),
        ]
    )

def ready_condition(obj: V1StatefulSet) -> bool:
    pods = get_pods_for_statefulset(obj.metadata.name, obj.metadata.namespace)
    ready = sum(pod_ready(pod) for pod in pods)
    total = obj.spec.replicas
    logger.info(f"{obj.metadata.name} ready={ready} total={total} ({(ready/total)*100:.0f}%)")
    return ready / total > 0.75

await wait_for_rollout(nodes, api_client, timeout=3000, condition=ready_condition)
```

src/deployments/core/kube_utils.py
PearsonWhite: Addressed in 73bc65a3b81c1ce49aa5b871544e84b4a614a9ea
See https://github.com/vacp2p/10ksim/pull/260#issuecomment-4390738591

src/deployments/core/kube_utils.py
PearsonWhite: Addressed in 2570e593ee6094009fb4cd17e58e02fc9eb4772e

AlbertoSoutullo (approved):

PR #262, opened 2026-05-05
global _api_client

PearsonWhite:
Use global api_client instead of passing it around

### DIFF
--- a/deployment.py
+++ b/deployment.py
@@ -7,12 +7,10 @@ from datetime import datetime
 from pathlib import Path
 from typing import Optional
 
-from kubernetes import config
-from kubernetes.client import ApiClient
 from ruamel import yaml
 
 # Project Imports
-from src.deployments.core.kube_utils import init_logger, set_config_file
+from src.deployments.core.kube_utils import init_config_file, init_logger
 from src.deployments.registry import registry as experiment_registry
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


AlbertoSoutullo (changes requested):
Hmm I dontt really agree with this approach.

First, there are a lot of places where api_client is still being used, which makes the PR a bit weird to review.
Then, if we want to simplify the usage of `api_client`I would not go with global.
Create something like a session:
```python
@dataclass
class KubeSession:
    kube_config: str
    api_client: client.ApiClient

    @classmethod
    def from_config(cls, kube_config: str) -> "KubeSession":
        config.load_kube_config(config_file=kube_config)
        return cls(kube_config=kube_config, api_client=client.ApiClient())

    def close(self) -> None:
        self.api_client.close()
```

Then you can do something like:
```python
session = KubeSession.from_config(str(kube_config))
try:
    await experiment.run(session, args, values_yaml)
finally:
    session.close()
```
This allows you to pass session through experiment boundaries only:
```python
BaseExperiment.run(session, args, values_yaml)
BaseExperiment._run(session, ...)
RegressionNodes.run(session, ...)
```

Then inside `BaseExperiment.run`, you can store `self._api_client = session.api_client` for convenience, and remove the global entirely.
There are probably even better ways to do this, but definitely this global approach... I don't see the benefit of it.

PR #264, opened 2026-05-08
Fix missing messages again

PearsonWhite:
* Filter None from _get_peers_missing_messages return value

* Update WakuTracer query for sent messages

----

Somehow this was broken again. _get_peers_missing_messages_for_shard was returning None or a list that included None, which caused an exception later.

### DIFF
--- a/src/analysis/mesh_analysis/readers/tracers/waku_tracer.py
+++ b/src/analysis/mesh_analysis/readers/tracers/waku_tracer.py
@@ -64,7 +64,7 @@ class WakuTracer(MessageTracer):
                     convert=self._trace_mixnet_in_logs,
                 ),
             ],
-            query="(sent relay message OR publishWithConn)",
+            query="sent relay message",
         )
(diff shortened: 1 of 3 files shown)


src/analysis/mesh_analysis/readers/tracers/waku_tracer.py
```diff
@@ -64,7 +64,7 @@ def with_sent_pattern_group(self) -> Self:
                     convert=self._trace_mixnet_in_logs,
                 ),
             ],
-            query="sent relay message",
+            query="(sent relay message OR publishWithConn)",
```
AlbertoSoutullo: Uh are you sure this is needed? Because this was with lightpush right?

PearsonWhite: Undid this change in f05f37399eab68f54cf3dfea61b056b8245cd78d

AlbertoSoutullo (approved):

PR #267, opened 2026-05-14
Connection Manager Experiment Deployment

### LINKED ISSUE #252: Connection manager
[Roadmap Link](https://roadmap.vac.dev/dst/ift/2026q2-libp2p-evaluation#connection-manager)

#### Description
Switch builder now has an option for pruning behavior, and connection manager has ephemeral tag mechanisms. More info:
https://github.com/vacp2p/nim-libp2p/blob/master/examples/tutorial_5_connmanager.nim
Help validating cahnges of watermark mode as well as connection scoring mechanism.
Provide feedback if examples are not clear, and create scenarios to validate the behavior and impact of this new changes.

radiken:
New experiment class with runs A–G and scale variants or evaluating nim-libp2p's connection manager watermark, grace period, protection, and oscillation behavior.

### DIFF
--- a/src/deployments/experiments/libp2p/connmanager.py
+++ b/src/deployments/experiments/libp2p/connmanager.py
@@ -181,14 +181,9 @@ class ConnManagerExperiment(BaseExperiment, BaseModel):
     # Group B peers: they dial hub (no bonus)
     # -------------------------------------------------------------------------
     async def _run_a(self, api_client, args, stack, config, namespace, image, workdir):
-        hub_svc, hub, governance_svc, peers_a, peers_b = self._build_run_a(config, namespace, image)
+        hub, peers_a, peers_b = self._build_run_a(config, namespace, image)
 
-        for name, obj in [
-            ("hub-svc", hub_svc),
-            ("nimp2p-service", governance_svc),
-        ]:
-            self.dump_yaml(obj, workdir, name)
-            await self.deploy(api_client, stack, args, {}, deployment=obj)
+        await self._deploy_services(namespace, workdir, api_client, stack, args)
 
         self.dump_yaml(hub, workdir, "hub")
         await self.deploy(api_client, stack, args, {}, deployment=hub, wait_for_ready=True)
@@ -205,16 +200,6 @@ class ConnManagerExperiment(BaseExperiment, BaseModel):
 
     def _build_run_a(self, config, namespace, image):
         outbound_peers = _outbound_peers_str(config.num_peers_outbound, namespace)
-
-        hub_svc = (
-            ServiceBuilder()
-            .with_name("hub")
-            .with_namespace(namespace)
-            .with_selector("app", "zerotenkay")
-            .with_selector("role", "hub")
-            .with_port(V1ServicePort(name="p2p", port=5000, target_port=5000))
-            .build()
-        )
 
         hub = (
             Libp2pStatefulSetBuilder()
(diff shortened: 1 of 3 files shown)


src/deployments/experiments/libp2p/connmanager.py
```diff
@@ -0,0 +1,969 @@
...
+                )
+                .with_image(image)
+                .with_label("role", "protected")
+                .with_option("NODE_ROLE", "RolePeer")
+                .with_option("PORT", "5000")
+                .with_option("DIAL_OUT", "true")
+                .with_option("HUB_ADDRS", _hub_addrs(config.num_hubs, namespace))
+                .with_option("PRIVATE_KEY", config.protected_peer_keys[0])
+                .build()
+            )
```
AlbertoSoutullo: This is weird. First you generate `protected`, and then if you have keys you generate `protected` again?
This can be simplified. What worries me is that it looks like you are just using the same key for every pod...
`config.protected_peer_keys[0]`

src/deployments/experiments/libp2p/connmanager.py
```diff
@@ -0,0 +1,969 @@
...
+                .with_namespace(namespace)
+                .with_selector("app", "zerotenkay")
+                .with_selector("role", "hub")
+                .with_port(V1ServicePort(name="p2p", port=5000, target_port=5000))
+                .build(),
+            ),
+            ("nimp2p-service", _governance_service(namespace)),
+        ]:
+            self.dump_yaml(obj, workdir, name)
+            await self.deploy(api_client, stack, args, {}, deployment=obj)
```
AlbertoSoutullo: This is literally pasted in a lot of different places. Extract it to avoid duplicated code.

AlbertoSoutullo (changes requested):

src/deployments/experiments/libp2p/connmanager.py
radiken: Refactored now, thanks!

src/deployments/experiments/libp2p/connmanager.py
radiken: Thanks for catching this, this is bad leftover from earlier iterations. Fixed this now and validated that it does not affect the test result as this is only for run C and we have only one protected node there.

AlbertoSoutullo (approved):
LGTM

src/deployments/libp2p/builders/builders.py
```diff
@@ -56,5 +56,5 @@ def with_image(self, image: Image) -> Self:
 
     def with_readiness_probe(self, probe: V1Probe) -> Self:
         container = find_libp2p_container_config(self.config)
-        container.with_readiness_probe(probe)
+        container.with_readiness_probe(probe, overwrite=True)
```
PearsonWhite, line 59:
```suggestion
        container.with_readiness_probe(probe)
```

PearsonWhite (changes requested):
Looks good. Extra changes can be eliminated by rebasing onto `master` after https://github.com/vacp2p/10ksim/pull/268 or grabbing e9562c0b4f4758547828267bc0edc43a41f9e5a5. Not a big deal, just makes it easier.

src/deployments/libp2p/builders/builders.py
radiken: Currently the experiment doesn't work without this line, but I see you fixed this in #269, so I am happy to remove this here

PearsonWhite: I got my changes mixed up. You can add that back in if you want. Just ping me and I'll re-approve.

PearsonWhite (approved):

src/deployments/libp2p/builders/builders.py
radiken: Added it back. Looks like I can still merge

PR #272, opened 2026-05-22
Allow to specify service + dns searchs

AlbertoSoutullo:


src/deployments/libp2p/builders/builders.py
```diff
@@ -33,13 +35,15 @@ def with_libp2p_config(
         self,
         name: str,
         namespace: str,
+        service: str,
```
radiken: This breaks the existing calls for `with_libp2p_config`, needs to go over all of them and add this param

src/deployments/libp2p/builders/nodes.py
radiken: missing args `namespace` and `dns_searches` for `Nodes.create_stateful_set_spec_config`

radiken (changes requested):

radiken (approved):
LGTM

PR #275, opened 2026-05-26
Add nimp2p-service to nimlibp2p nodes

PearsonWhite:
Change on top of #272

### DIFF
--- a/src/deployments/libp2p/builders/builders.py
+++ b/src/deployments/libp2p/builders/builders.py
@@ -1,5 +1,6 @@
 # Python Imports
-from typing import List, Self
+from typing import List, Literal, Self
+
 from kubernetes.client import V1EnvVar, V1Probe, V1StatefulSet
 from pydantic import PositiveInt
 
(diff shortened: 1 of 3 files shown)


src/deployments/libp2p/builders/nodes.py
```diff
@@ -47,15 +47,16 @@ def create_pod_spec_config(
         dns_searches: List[str],
         namespace: str,
     ) -> PodSpecConfig:
-        dns_searches_complete = (
-            [f"{service}.{namespace}.svc.cluster.local" for service in dns_searches]
-            if dns_searches
-            else None
-        )
+        dns_config = None
```
PearsonWhite, line 50: This avoids adding `dnsConfig: {}` to pod_spec_configs when `dns_searches` is not specified (`None`).

src/deployments/libp2p/builders/builders.py
```diff
@@ -1,5 +1,6 @@
 # Python Imports
-from typing import List, Self
+from typing import List, Literal, Self
```
PearsonWhite, line 2: Add missing import.

AlbertoSoutullo (approved):

PR #276, opened 2026-05-26
Shadow simulator: run GossipSub experiments

### LINKED ISSUE #220: Shadow
[Roadmap Link](https://roadmap.vac.dev/dst/ift/2026q2-dst-tooling#shadow)

#### Description
Shadow simulator is hardcoded at the moment in dst-libp2p-test-node for the nim version.
We want to abstract builders from the main repository to take into account shadow.
We want to be able to launch experiments on K8s, deciding if we want to use a realistic environment, or
shadow.
This needs to take into account further data analysis of the main pipeline. Can probably be split into
subtasks.

radiken:
Runs N nim libp2p peers + a publisher inside the Shadow simulator on one k8s pod. `python deployment.py shadow-gossipsub` (or `multi-shadow-gossipsub` to scale 10/30/100).

- `shadow/builders.py`: renders `shadow.yaml` + the ConfigMap, PVC, and runner Job (init container fetches the dynamic-linked test-node binary; main container `radiken/dst-shadow-base`, seccomp Unconfined + SYS_PTRACE).
- `shadow/runtime.py`: waits for the Job, then copies output off a PVC via a short-lived reader pod (`kubectl cp`) — keeps per-host logs/metrics off pod stdout (~18x less at 100 peers) — and flattens per-host logs to `logs/<peer>.log` so the existing `FileStack` reads them unchanged.
- `experiments/libp2p/shadow_gossipsub.py` + `multi_shadow_gossipsub.py`: the experiment + scale driver.
- `kube_utils.py`: ConfigMap + PVC added to cleanup dispatch. `.gitignore`: `out/*` -> `out/`.

Metrics analysis is in #285.


src/deployments/shadow/runtime.py
```diff
@@ -0,0 +1,151 @@
...
+    job_name: str,
+    pvc_name: str,
+    reader_image: str,
+    dest_dir: Path,
+    node_pin: Optional[str] = None,
+    reader_ready_timeout_s: int = 120,
+) -> None:
+    """Pull Shadow's output into `dest_dir`: `shadow_stdout.log` (Job pod stdout via
+    kubectl logs), `shadow_data/` (the PVC's shadow.data, copied out via a reader pod +
+    kubectl cp), and `logs/<host>.log` (flattened for FileStack)."""
```
AlbertoSoutullo, line 118: Hmm isn't this ignoring the kubeconfig that can be optionally passed in deployment.py?

radiken: Good catch! Fixed by adding a `get_config_file()` helper in `core/kube_utils.py` and routing both calls through it.

AlbertoSoutullo (approved):
LGTM

PR #269, opened 2026-05-15
BaseExperiment: Moved run params into fields

PearsonWhite:
* Move api_client, ExitStack, args (skip-check, etc.), to BaseExperiment fields

* Add namespace and config as default BaseExperiment field

* Derive workdir from output_folder

* Setup values_yaml and experiment config outside of run

* Setup output paths inside of BaseExperiment class

* Move metrics readiness probe out of default nimlibp2p node builder

PR #277, opened 2026-05-27
multi_experiment fix

radiken:
## Summary

PR #269 (`BaseExperiment: moved run params into fields`) simplified `BaseExperiment.run()` to zero-arg, but `Multiple._run()` was not updated. 

Effect: `Multiple` subclass (including the existing `multi_nimlibp2p`) raises at startup with:

```
TypeError: Multiple._run() missing 5 required positional arguments:
'api_client', 'workdir', 'args', 'values_yaml', and 'stack'
```

This PR refactors `Multiple._run()` to the new shape: reads `api_client` / `namespace` / `skip_check` / `dry_run` / `output_folder` / `config` from `self`, and constructs + dispatches each child with the same kwargs that `deployment.py` uses.

### DIFF
--- a/src/deployments/experiments/multi_experiment.py
+++ b/src/deployments/experiments/multi_experiment.py
@@ -3,18 +3,12 @@ import logging
 import random
 import traceback
 from abc import abstractmethod
-from argparse import Namespace
-from contextlib import ExitStack
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


AlbertoSoutullo (approved):
Nice change, thanks a lot.

PR #278, opened 2026-05-28
Add dump to BaseExp and fix run defs

PearsonWhite:
Dump yaml in BaseExperiment

Fix _run defs in experiment classes

### DIFF
--- a/src/deployments/helm_deployment/nimlibp2p/experiments/mix/multiple_regression.py
+++ b/src/deployments/helm_deployment/nimlibp2p/experiments/mix/multiple_regression.py
@@ -1,18 +1,14 @@
 # Python Imports
 import logging
 import random
-from argparse import Namespace
 from asyncio import sleep
-from contextlib import ExitStack
 from copy import deepcopy
 from datetime import datetime
 from datetime import timezone as dt_timezone
 from pathlib import Path
-from typing import List, Optional
+from typing import List
 
-from kubernetes.client import ApiClient
 from pydantic import BaseModel, ConfigDict
-from ruamel import yaml
 
 # Project Imports
 from src.deployments.core.kube_utils import dict_get, dict_set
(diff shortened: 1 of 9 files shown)


AlbertoSoutullo (dismissed):
`_run_d` is not updated.

src/deployments/helm_deployment/nimlibp2p/experiments/mix/multiple_regression.py
```diff
@@ -61,13 +52,13 @@ async def _run(
...
                 / image
                 / f"delay_{delay}_jitter_{jitter}_messages_{messages}_mix_{mix_nodes}_gossip_{nonmix_nodes}_rand_{random_number}"
             )
-            exp_args = deepcopy(args)
-            exp_args.workdir = exp_workdir
-            exp_args.delay = None
+            exp_config = deepcopy(self.config)
+            exp_config.workdir = exp_workdir
+            exp_config.delay = None
             exp_values_yaml = deepcopy(values_yaml)
```
radiken, line 62: `values_yaml` is dropped from the params, so I think this wouldn't work?

src/deployments/helm_deployment/nimlibp2p/experiments/mix/multiple_regression.py
```diff
@@ -129,7 +120,7 @@ async def _run(
 
             experiment = NimMixNodes()
             logger.info(f"running wtih delay {delay} jitter {jitter}")
-            experiment.run(api_client, exp_args, exp_values_yaml)
+            experiment.run(api_client, exp_config, exp_values_yaml)
```
radiken, line 123: Same here, `api_client` is not defined

radiken (changes requested):

src/deployments/helm_deployment/nimlibp2p/experiments/mix/multiple_regression.py
PearsonWhite: I think this whole test is deprecated at this point. I'll have to clean up some of this old stuff in a later PR.

src/deployments/helm_deployment/nimlibp2p/experiments/mix/multiple_regression.py
PearsonWhite: Same as before. I don't think it's worth fixing the helm deployment experiments right now.

radiken (approved):
Just have to make the lint ci happy

PR #285, opened 2026-05-28
Shadow: auto-fired post-run analysis (metrics + message reliability)

### LINKED ISSUE #220: Shadow

radiken:
Makes a Shadow run analyze itself like a k8s run, mirroring the connmanager experiment (deploy + analysis + plots in one `deployment.py` invocation, defined in the experiment).

**Depends on #276**

- `analysis/metrics/shadow_metrics.py`: split each peer's `/metrics` snapshots, import into a throwaway dockerized VictoriaMetrics (synthesized timestamps + k8s-style labels), run the existing `Scrapper` -> same per-metric CSVs.
- `shadow_gossipsub._run_analysis()`: fires after the run (best-effort, never fails it) -> bandwidth CSVs (above) + message reliability via `Nimlibp2pAnalyzer` on the flattened `logs/` (`FileStack`).
- `scrapper.py`: `kube_config` made optional (unused port-forward path).

Needs Docker locally for the ephemeral VM.

### DIFF
--- a/src/analysis/metrics/scrapper.py
+++ b/src/analysis/metrics/scrapper.py
@@ -16,10 +16,10 @@ class Scrapper(BaseModel):
     _config: ScrapeConfig
     _k8s: object
 
-    def __init__(self, kube_config: str, config: ScrapeConfig):
+    def __init__(self, kube_config: Optional[str] = None, config: ScrapeConfig = None):
         self._url = config.url
         self._config = config
-        self._k8s = kubernetes_manager.KubernetesManager(kube_config)
+        self._k8s = kubernetes_manager.KubernetesManager(kube_config) if kube_config else None
 
     def query_and_dump_metrics(self):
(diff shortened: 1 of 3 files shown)


AlbertoSoutullo (commented):
II would say that `_run_analysis()` should be invoked after `_run` finishes, because then it is the cleanup phase.
I would move the analysis to a post cleanup phase to avoid potential issues and a more stable flow.

AlbertoSoutullo (approved):

PR #286, opened 2026-06-08
kube utils tests

radiken:
- Set up testing CI
- Add `kube_rollout` and `kube_cleanup` tests

### DIFF
--- a/src/deployments/core/k8s_cleanup.py
+++ b/src/deployments/core/k8s_cleanup.py
@@ -288,7 +288,7 @@ def poll_namespace_has_objects(
         "DaemonSet": lambda: apps_v1.list_namespaced_daemon_set(namespace).items,
         "Job": lambda: batch_v1.list_namespaced_job(namespace).items,
         "ReplicationController": lambda: v1.list_namespaced_replication_controller(namespace).items,
-        "CronJob": lambda: batch_v1.list_namespaced_cron_job(namespace),
+        "CronJob": lambda: batch_v1.list_namespaced_cron_job(namespace).items,
     }
 
     for type, check in resource_checks.items():
--- a/src/deployments/core/k8s_cleanup.py
+++ b/src/deployments/core/k8s_cleanup.py
@@ -272,7 +272,7 @@ def poll_namespace_has_objects(
     """
     types = types if types else ["Deployment", "StatefulSet", "DaemonSet", "ReplicaSet", "Pod"]
     logger.debug(f"Checking in namespace `{namespace}` for types: `{types}`")
-    v1 = client.CoreV1Api()
+    v1 = client.CoreV1Api(api_client)
     apps_v1 = client.AppsV1Api(api_client)
     batch_v1 = client.BatchV1Api(api_client)
 
--- a/src/deployments/core/tests/test_k8s_cleanup.py
+++ b/src/deployments/core/tests/test_k8s_cleanup.py
@@ -1,0 +1,119 @@
+from unittest.mock import MagicMock
+
+import pytest
+from kubernetes.client.rest import ApiException
+
+from src.deployments.core import k8s_cleanup
+from src.deployments.core.k8s_cleanup import (
+    cleanup_resources,
+    poll_namespace_has_objects,
+    wait_for_no_objs_in_namespace,
+)
+
+
+def _list_result(items):
+    r = MagicMock()
+    r.items = items
+    return r
+
+
+# --------------------------------------------------------------------------- #
+# poll_namespace_has_objects  (lists resources via k8s client)
+# --------------------------------------------------------------------------- #
+class TestPollNamespaceHasObjects:
+    def _patch_apis(self, mocker, *, core=None, apps=None, batch=None):
+        core = core or MagicMock()
+        apps = apps or MagicMock()
+        batch = batch or MagicMock()
+        mocker.patch.object(k8s_cleanup.client, "CoreV1Api", return_value=core)
+        mocker.patch.object(k8s_cleanup.client, "AppsV1Api", return_value=apps)
+        mocker.patch.object(k8s_cleanup.client, "BatchV1Api", return_value=batch)
+        return core, apps
+
+    def test_true_when_a_statefulset_exists(self, mocker):
+        core, apps = self._patch_apis(mocker)
+        core.list_namespaced_pod.return_value = _list_result([])
+        apps.list_namespaced_deployment.return_value = _list_result([])
+        apps.list_namespaced_stateful_set.return_value = _list_result([object()])
+        apps.list_namespaced_daemon_set.return_value = _list_result([])
+        apps.list_namespaced_replica_set.return_value = _list_result([])
+
+        assert poll_namespace_has_objects("ns", MagicMock()) is True
+
+    def test_false_when_namespace_empty(self, mocker):
+        core, apps = self._patch_apis(mocker)
+        core.list_namespaced_pod.return_value = _list_result([])
+        apps.list_namespaced_deployment.return_value = _list_result([])
+        apps.list_namespaced_stateful_set.return_value = _list_result([])
+        apps.list_namespaced_daemon_set.return_value = _list_result([])
+        apps.list_namespaced_replica_set.return_value = _list_result([])
+
+        assert poll_namespace_has_objects("ns", MagicMock()) is False
+
+    def test_api_exception_is_swallowed(self, mocker):
+        core, _ = self._patch_apis(mocker)
+        core.list_namespaced_pod.side_effect = ApiException()
+        assert poll_namespace_has_objects("ns", MagicMock(), types=["Pod"]) is False
+
+
+# --------------------------------------------------------------------------- #
+# wait_for_no_objs_in_namespace  (loop around poll_namespace_has_objects)
+# --------------------------------------------------------------------------- #
+class TestWaitForNoObjsInNamespace:
+    def test_returns_when_namespace_clears(self, mocker):
+        poll = mocker.patch.object(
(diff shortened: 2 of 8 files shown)


AlbertoSoutullo (commented):
Thanks! Please, move the removal of outdated tests into a separate PR and we merge that one first

AlbertoSoutullo (changes requested):
This file is the perfect example of the cleanup we need to do:
In this single file we are mixing:
- rollout logic
- cleanup
- YAML/Helm helpers
- dict utilities 
- time utilities
So before start adding tests to just a part of this file, we need to correctly split this.
K8s related things can be split by modules: `k8s_rollout.py`, `k8s_cleanup.py` that can remain in `src/deployment/core` for example.

Then, the general utils  like `yaml_utils.py`, `dict_utils.py` or `time_utils.py` can go to `src/utils` because they are generic utils that can be used in the whole repo.

src/deployments/core/tests/test_k8s_cleanup.py
```diff
@@ -0,0 +1,119 @@
...
+        apps.list_namespaced_stateful_set.return_value = _list_result([])
+        apps.list_namespaced_daemon_set.return_value = _list_result([])
+        apps.list_namespaced_replica_set.return_value = _list_result([])
+
+        assert poll_namespace_has_objects("ns", MagicMock()) is False
+
+    def test_api_exception_is_swallowed(self, mocker):
+        core, _ = self._patch_apis(mocker)
+        core.list_namespaced_pod.side_effect = ApiException()
+        assert poll_namespace_has_objects("ns", MagicMock(), types=["Pod"]) is False
```
AlbertoSoutullo: Does not prove exceptions are swallowed for all resource types.

src/deployments/core/tests/test_k8s_cleanup.py
```diff
@@ -0,0 +1,119 @@
...
+class TestCleanupResources:
+    def test_deletes_each_named_statefulset(self, mocker):
+        apps = MagicMock()
+        mocker.patch.object(k8s_cleanup.client, "AppsV1Api", return_value=apps)
+
+        cleanup_resources({"StatefulSet": ["nodes-0", "nodes-1"]}, "ns", api_client=MagicMock())
+
+        apps.delete_namespaced_stateful_set.assert_any_call("nodes-0", "ns")
+        apps.delete_namespaced_stateful_set.assert_any_call("nodes-1", "ns")
+        assert apps.delete_namespaced_stateful_set.call_count == 2
```
AlbertoSoutullo, line 126: This checks the calls, but not that unsupported/missing kinds are ignored safely

src/deployments/core/tests/test_k8s_cleanup.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def test_false_when_namespace_empty(self, mocker):
+        core, apps = self._patch_apis(mocker)
+        core.list_namespaced_pod.return_value = _list_result([])
+        apps.list_namespaced_deployment.return_value = _list_result([])
+        apps.list_namespaced_stateful_set.return_value = _list_result([])
+        apps.list_namespaced_daemon_set.return_value = _list_result([])
+        apps.list_namespaced_replica_set.return_value = _list_result([])
+
+        assert poll_namespace_has_objects("ns", MagicMock()) is False
```
AlbertoSoutullo: This test is fine because `poll_namespace_has_objects()` defaults to this.
If later someone changes the default types to include "Service" or "Job", this test may fail with an unconfigured MagicMock, or worse, pass/fail for the wrong reason.
We should make it more explicit:
```python
assert poll_namespace_has_objects(
    "ns",
    MagicMock(),
    types=["Deployment", "StatefulSet", "DaemonSet", "ReplicaSet", "Pod"],
) is False
```
That way the test says: “given these resource types are empty, return false.” It no longer depends silently on the function’s default list.

src/deployments/core/tests/test_k8s_cleanup.py
AlbertoSoutullo, line 1: The tests do not assert API constructors receive `api_client` consistently. This matters because `poll_namespace_has_objects` currently does client.CoreV1Api() without passing api_client.

In the `poll_namespace_has_objects` implementation we have:
```python
v1 = client.CoreV1Api()
apps_v1 = client.AppsV1Api(api_client)
batch_v1 = client.BatchV1Api(api_client)
```

`AppsV1Api` and `BatchV1Api` use the passed `api_client`, but `CoreV1Api` does not.
So if the caller passes a custom Kubernetes client/session/config, core resources like Pods, Services, PVCs, and ConfigMaps may be listed using the default global client instead.
These tests tests don’t catch this because they patch:
`mocker.patch.object(k8s_cleanup.client, "CoreV1Api", return_value=core)`
But this don't assert how it was called. A correct assertion would be:
```python
api_client = MagicMock()
core_api = mocker.patch.object(k8s_cleanup.client, "CoreV1Api", return_value=core)

poll_namespace_has_objects("ns", api_client, types=["Pod"])

core_api.assert_called_once_with(api_client)
```
Then, this test will fail because the code calls `client.CoreV1Api()` instead of `client.CoreV1Api(api_client)`.

So the `poll_namespace_has_objects` implementation should change to `v1 = client.CoreV1Api(api_client)`.
cc: @PearsonWhite @mamoutou-diarra

AlbertoSoutullo (changes requested):

src/deployments/core/tests/test_k8s_cleanup.py
radiken: Good point! Parametrized it over all the resource types now

src/deployments/core/tests/test_k8s_cleanup.py
radiken: Added `test_unsupported_or_missing_kinds_are_ignored` for this

src/deployments/core/tests/test_k8s_cleanup.py
radiken: agree, updated

src/deployments/core/tests/test_k8s_cleanup.py
radiken: Good catch, fixed it and confirmed it fails on the old code and passes with the fix

src/deployments/core/tests/test_k8s_rollout.py
```diff
@@ -0,0 +1,170 @@
...
+        assert check_pod_condition(pod) is False
+
+    def test_custom_tuple_condition(self):
+        pod = _pod_with_conditions([("Initialized", "True")])
+        assert check_pod_condition(pod, ("Initialized", "True")) is True
+        assert check_pod_condition(pod, ("Ready", "True")) is False
+
+    def test_callable_condition_is_delegated_to(self):
+        pod = MagicMock()
+        assert check_pod_condition(pod, lambda p: True) is True
```
AlbertoSoutullo: This can be improved with:
```python
def test_callable_condition_false_is_propagated(self):
    pod = MagicMock()
    condition = MagicMock(return_value=False)
    assert check_pod_condition(pod, condition) is False
    condition.assert_called_once_with(pod)
```

So we verify the result is respected and the callabe received the pod object.

src/deployments/core/tests/test_k8s_rollout.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+# --------------------------------------------------------------------------- #
+# poll_rollout_status  (reads live object via get_namespaced)
+# --------------------------------------------------------------------------- #
+class TestPollRolloutStatus:
+    def test_statefulset_all_replicas_ready_is_true(self, mocker):
+        obj = _statefulset(desired=3, ready=3)
+        mocker.patch.object(k8s_rollout, "get_namespaced", return_value=obj)
+        assert poll_rollout_status(obj) is True
```
AlbertoSoutullo: The function assumes that `poll_rollout_status()` fetches the latest object from Kubernetes via `get_namespaced(...)` and then evaluates `live_obj`. But because the object we pass in and the object returned by the mock is literally the same the test cannot tell whether the implementation uses the fresh object or the original one.

This could be improved with:
```python
def test_statefulset_uses_live_object(self, mocker):
    stale = _statefulset(desired=3, ready=0)

    live = _statefulset(desired=3, ready=3)

    mocker.patch.object(
        k8s_rollout,
        "get_namespaced",
        return_value=live,
    )

    assert poll_rollout_status(stale) is True
```

Maybe it would be nice to add the inverse as well:
```python
def test_statefulset_uses_live_object_not_stale_object(self, mocker):
    stale = _statefulset(desired=3, ready=3)

    live = _statefulset(desired=3, ready=1)

    mocker.patch.object(
        k8s_rollout,
        "get_namespaced",
        return_value=live,
    )

    assert poll_rollout_status(stale) is False
```

AlbertoSoutullo (changes requested):

src/deployments/core/tests/test_k8s_rollout.py
radiken: Thanks, applied

src/deployments/core/tests/test_k8s_rollout.py
radiken: agree, applied

src/deployments/core/k8s_cleanup.py
AlbertoSoutullo: This should be `batch_v1.list_namespaced_cron_job(namespace).items`

AlbertoSoutullo (changes requested):
One last thing

src/deployments/core/k8s_cleanup.py
radiken: Good catch, thanks!

AlbertoSoutullo (approved):
LGTM. Thanks!

PR #287, opened 2026-06-08
Fix poll_rollout_status returning a tuple for Deployments

radiken:
`poll_rollout_status` returns `(available, desired)` instead of a bool. This shoud not be intended, now `wait_for_rollout` treated any deployment as ready immediately even at 0 available replicas. Fixing it to return `available == desired`.


AlbertoSoutullo (approved):
Thanks, I missed when this was changed. Weird.

PR #289, opened 2026-06-08
Add security_context to ContainerConfig

### LINKED ISSUE #153: General tooling (recurring)

#### Description

PearsonWhite:
Fixes with_network_delay which relies on initContainer having NET_ADMIN


AlbertoSoutullo (approved):
It would be nice to also have an example of the config that you pass to `build_container` to use this

PR #290, opened 2026-06-10
Split kube_utils into focused modules

radiken:
Follow-up to the review on #286: `kube_utils.py` had grown into a catch-all — rollout logic, cleanup, kubectl apply, YAML/Helm helpers, dict utils and time utils all living in one ~1.3k-line file. This splits it up so each concern has its own module, before we start adding tests on top.


src/deployments/core/k8s_kubeconfig.py
AlbertoSoutullo, line 1: Maybe I would call this `k8s_kubeconfig.py` or `k8s_config_file.py` maybe.

src/utils/cli_utils.py
```diff
@@ -0,0 +1,27 @@
...
+        matches = re.search(f"--{flag}=(?P<numMessages>\\d+)", node)
+        try:
+            return int(matches["numMessages"])
+        except (TypeError, IndexError):
+            pass
+    return None
+
+
+def assert_equals(obj_1, obj_2):
+    assert obj_1 == obj_2, f"Assertion failed: `{obj_1}` == `{obj_2}`"
```
AlbertoSoutullo: Is this being used?

AlbertoSoutullo (commented):
I think there are some other improvements to do but this can be in another PR. It is a nice split. Slowly improving.

src/deployments/core/k8s_kubeconfig.py
radiken: updated to `k8s_kubeconfig.py`

src/utils/cli_utils.py
radiken: Good catch, it's not used anywhere, removed

AlbertoSoutullo (approved):
Thanks!

PR #291, opened 2026-06-10
Improvements for deployment building

### LINKED ISSUE #153: General tooling (recurring)

PearsonWhite:
* Added service_account_name and security_context to builders
* Make helpers.py into a folder for core/configs
* Fix PodTemplateSpecConfig default factory
* Fix with_volume_mount and with_port for ContainerConfig
* Added automount_service_account_token
* Add namespace check to StatefulSet builder
* Fix dns_config.searches condition


AlbertoSoutullo (commented):
It would be nice that if we add new functions they already go with tests.

src/deployments/core/k8s_cleanup.py
```diff
@@ -30,6 +30,10 @@ def get_cleanup_resources(yamls: List[yaml.YAMLObject], types: Optional[List[str
         "Service": [],
         "ConfigMap": [],
         "PersistentVolumeClaim": [],
+        "Role": [],
+        "RoleBinding": [],
+        "ConfigMap": [],
```
AlbertoSoutullo: `ConfigMap` is duplicated

AlbertoSoutullo (changes requested):

src/deployments/core/k8s_cleanup.py
PearsonWhite: Fixed in 3d2025c19087225e5f49d7bb4d643361aed51a96

AlbertoSoutullo (approved):

PR #292, opened 2026-06-11
logoscore integration

### LINKED ISSUE #228: Logos-core evaluation
[Roadmap Link](https://roadmap.vac.dev/dst/logos/2026q2-logos-core-evaluation#logos-core-evaluation)

#### Description
Integrate logos-core in the DST lab, evaluating its behaviour.
Prepare new tools if needed, and report performance or issues if they are found.

PearsonWhite:
Should go in after https://github.com/vacp2p/10ksim/pull/291

### DIFF
--- a/src/deployments/core/configs/helpers/identity.py
+++ b/src/deployments/core/configs/helpers/identity.py
@@ -3,21 +3,32 @@ from typing import Optional
 
 # Project Imports
 from src.deployments.core.configs.helpers.utils import HigherConfigTypes
-from src.deployments.core.configs.pod import PodTemplateSpecConfig
+from src.deployments.core.configs.pod import PodConfig, PodTemplateSpecConfig
 from src.deployments.core.configs.statefulset import StatefulSetConfig, StatefulSetSpecConfig
 
 
-def apply_pod_template_spec_config(config: PodTemplateSpecConfig, namespace: str, app: str):
-    config.name = "pod"
+def apply_pod_config(config: PodConfig, namespace: str, name: str, app: str):
+    config.name = name
     config.namespace = namespace
     if app:
         config.with_app(app)
 
 
-def apply_stateful_set_spec_config(config: StatefulSetSpecConfig, namespace: str, app: str):
+def apply_pod_template_spec_config(
+    config: PodTemplateSpecConfig, namespace: str, name: str, app: str
+):
+    config.name = name
+    config.namespace = namespace
     if app:
         config.with_app(app)
-    apply_pod_template_spec_config(config.pod_template_spec_config, namespace, app)
+
+
+def apply_stateful_set_spec_config(
+    config: StatefulSetSpecConfig, namespace: str, name: str, app: str
+):
+    if app:
+        config.with_app(app)
+    apply_pod_template_spec_config(config.pod_template_spec_config, namespace, name, app)
 
 
 def apply_stateful_set_config(
@@ -25,15 +36,17 @@ def apply_stateful_set_config(
 ):
     config.name = name
     config.namespace = namespace
-    apply_stateful_set_config(config.stateful_set_spec, namespace, app)
+    apply_stateful_set_spec_config(config.stateful_set_spec, namespace, name, app)
 
 
 def apply_identity(config: HigherConfigTypes, name: str, namespace: str, app: Optional[str] = None):
     if isinstance(config, StatefulSetConfig):
-        return apply_stateful_set_config(config, namespace, name, app)
+        return apply_stateful_set_config(config, namespace=namespace, name=name, app=app)
     if isinstance(config, StatefulSetSpecConfig):
-        return apply_stateful_set_spec_config(config, namespace, app)
+        return apply_stateful_set_spec_config(config, namespace=namespace, name=name, app=app)
+    if isinstance(config, PodConfig):
+        return apply_pod_config(config, namespace=namespace, name=name, app=app)
     elif isinstance(config, PodTemplateSpecConfig):
-        return apply_pod_template_spec_config(config, namespace, app)
+        return apply_pod_template_spec_config(config, namespace=namespace, app=app)
 
     raise ValueError(f"Unknown config type: {type(config)}")
--- a/src/deployments/core/configs/helpers/identity.py
+++ b/src/deployments/core/configs/helpers/identity.py
@@ -11,7 +11,7 @@ def apply_pod_config(config: PodConfig, namespace: str, name: str, app: str):
     config.name = name
     config.namespace = namespace
     if app:
-        config.with_app(app)
+        config.with_app(app, overwrite=True)
 
 
 def apply_pod_template_spec_config(
@@ -20,14 +20,14 @@ def apply_pod_template_spec_config(
     config.name = name
     config.namespace = namespace
     if app:
-        config.with_app(app)
+        config.with_app(app, overwrite=True)
 
 
 def apply_stateful_set_spec_config(
     config: StatefulSetSpecConfig, namespace: str, name: str, app: str
 ):
     if app:
-        config.with_app(app)
+        config.with_app(app, overwrite=True)
     apply_pod_template_spec_config(config.pod_template_spec_config, namespace, name, app)
 
 
--- a/src/deployments/core/configs/helpers/identity.py
+++ b/src/deployments/core/configs/helpers/identity.py
@@ -47,6 +47,6 @@ def apply_identity(config: HigherConfigTypes, name: str, namespace: str, app: Op
     if isinstance(config, PodConfig):
         return apply_pod_config(config, namespace=namespace, name=name, app=app)
     elif isinstance(config, PodTemplateSpecConfig):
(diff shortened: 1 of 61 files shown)


src/deployments/core/configs/helpers/identity.py
```diff
@@ -0,0 +1,52 @@
...
+
+def apply_identity(config: HigherConfigTypes, name: str, namespace: str, app: Optional[str] = None):
+    if isinstance(config, StatefulSetConfig):
+        return apply_stateful_set_config(config, namespace=namespace, name=name, app=app)
+    if isinstance(config, StatefulSetSpecConfig):
+        return apply_stateful_set_spec_config(config, namespace=namespace, name=name, app=app)
+    if isinstance(config, PodConfig):
+        return apply_pod_config(config, namespace=namespace, name=name, app=app)
+    elif isinstance(config, PodTemplateSpecConfig):
+        return apply_pod_template_spec_config(config, namespace=namespace, app=app)
```
AlbertoSoutullo: This is missing a required argument (`name`), is this working?

src/deployments/pod_api_requester/builder.py
AlbertoSoutullo: `PodSpecConfig` has no namespace field.

src/deployments/logos_core/builders/nodes.py
```diff
@@ -0,0 +1,131 @@
...
+# Project Imports
+from src.deployments.core.configs.container import ContainerConfig, Image
+from src.deployments.core.configs.helpers.identity import apply_identity
+from src.deployments.core.configs.helpers.utils import find_container_config, get_config
+from src.deployments.core.configs.pod import PodSpecConfig
+
+
+class NodesBuilder(StatefulSetBuilder):
+    _service_name: str = PrivateAttr(default="logos-core-service")
+    _service_account_name: str = PrivateAttr(default="secret-creator")
```
AlbertoSoutullo, line 29: Why is this `secred-creator` and then everywhere else is `secred-creator2`?
Lets not hardcode this type of things.

src/deployments/logos_core/builders/nodes.py
```diff
@@ -0,0 +1,131 @@
...
+        service = V1Service(
+            api_version="v1",
+            kind="Service",
+            metadata=V1ObjectMeta(
+                name="core-nodes-internal",
+                namespace=self._namespace,
+            ),
+            spec=V1ServiceSpec(
+                cluster_ip="None",
+                selector={"app": "zerotenkay-core"},
```
AlbertoSoutullo: Again, this is hardcoded to `zerotenkay-core` but the builder supports `with_app()`.

AlbertoSoutullo (changes requested):

src/deployments/core/configs/helpers/identity.py
PearsonWhite: Fixed in cbe12e4a1648c43394dddcb5308ae7ce0d5a6b3e

src/deployments/pod_api_requester/builder.py
PearsonWhite: Fixed in 6f8325a624ea2f4966a0d795f1c679e78f151d23

src/deployments/logos_core/builders/nodes.py
PearsonWhite: Fixed by 6f8325a624ea2f4966a0d795f1c679e78f151d23 and 1bad8d496748b1e99be7b600eeb0a912e58c7d1b

src/deployments/logos_core/builders/nodes.py
PearsonWhite: Fixed in c548bcd73f21ef671baeac84934c3c66dff01b67

src/deployments/logos_core/builders/request_builder.py
```diff
@@ -0,0 +1,258 @@
...
+        self._secret_creator_binding_name = binding_name
+        return self
+
+    def with_pod_service_reader_binding_name(self, binding_name: str) -> Self:
+        self._pod_service_reader_binding_name_logoscore = binding_name
+        return self
+
+    def with_service_name(self, service_name: str) -> Self:
+        self._service_name = service_name
+        return self
```
AlbertoSoutullo, line 132: This is prone to errors.
For example, in the experiment you do:
```python
publisher_builder = (
    LogoscorePodApiRequester()
    .with_namespace(self.namespace)
    .with_image(Image(repo="pearsonwhite/dst-lc-api", tag="1-amd"))
    .with_mode("server")
    .with_service_account_name(service_account_name)
    .with_service_name(_LOGOSCORE_PUBLISHER["service_name"])
    .with_logoscore_profile(
        self.namespace, name="logoscore-requester", app=_LOGOSCORE_PUBLISHER["app"]
    )
    .with_dns_search(f"{bootstrap_service}.{self.namespace}.svc.cluster.local")
    .with_dns_search(f"{relay_service}.{self.namespace}.svc.cluster.local")
)
```
`with_logoscore_profile()` works because you called first `with_service_account_name()`. If you call first `with_logoscore_profile`, this will access `self._service_account_name` which will not be updated yet.

src/deployments/logos_core/experiment.py
```diff
@@ -0,0 +1,388 @@
...
+            try:
+                await init_token(self.namespace, indexed_name, relay_service)
+            except Exception as e:
+                logger.error(f"e: {e}")
+            await init_node(self.namespace, indexed_name, relay_service, enrs)
+
+        for name, service in zip([bootstrap_name, relay_name], [bootstrap_service, relay_service]):
+            for index in range(self.config.num_nodes):
+                indexed_name = f"{name}-{index}"
+                await start_node(self.namespace, indexed_name, relay_service)
```
AlbertoSoutullo: Is this supposed to be always using `relay_service` for both bootstrap and relay? Because `subscribe()` uses variable `service`, and `send()` uses also variable `service.`

AlbertoSoutullo (changes requested):

src/deployments/logos_core/builders/request_builder.py
PearsonWhite: Fixed in 3ff7e452b6e27842d8fdeac47977e9e46f3dacd7

src/deployments/logos_core/experiment.py
PearsonWhite: Changed in 3ff7e452b6e27842d8fdeac47977e9e46f3dacd7

AlbertoSoutullo (changes requested):
We still have the same dependency issues with the builder as how it is implemented right now.
ie `apply_logoscore_profile`  does: 
`f"{self._service_name}.{self._namespace}.svc.cluster.local"`
If we don't do `with_service_name()` first, this will create "None.<namespace>.svc.cluster.local".

We could, from my pov:
1. Make the profile method take all required inputs
2. Reconcile config whenever these setters are called
3. Validate in build() / build_dependencies() that required state is present and consistent

src/deployments/logos_core/builders/request_builder.py
```diff
@@ -0,0 +1,254 @@
...
+    V1PolicyRule,
+    V1ResourceRequirements,
+    V1Role,
+    V1RoleBinding,
+    V1RoleRef,
+    V1Service,
+    V1ServiceAccount,
+    V1ServicePort,
+    V1ServiceSpec,
+    V1Subject,
```
radiken:
```suggestion
    RbacV1Subject,
```
This is renamed from kubernetes version 33

src/deployments/logos_core/builders/request_builder.py
```diff
@@ -0,0 +1,259 @@
...
+        return self
+
+    def with_pod_service_reader_binding_name(self, binding_name: str) -> Self:
+        self._pod_service_reader_binding_name_logoscore = binding_name
+        return self
+
+    def with_service_name(self, service_name: str) -> Self:
+        self._service_name = service_name
+        self._reconcile()
+        return self
```
AlbertoSoutullo, line 132: Why `_reconcile`is not called on every method?

PearsonWhite: `_reconcile` applies the settings relevant to the logoscore requester. Whenever a variable related to that is changed, we re-apply all the settings. That way, when one variable (like `namespace`) affects multiple settings, all of those settings are updated when that variable is changed.

AlbertoSoutullo: Still, `reconcile()` assumes `dns_config` already exists. So we still have the same issue.
`self.config.pod_spec_config.dns_config.searches.clear()` this only works if `dns_config` is already intialized, and this is called inside `_reconcile()`.

AlbertoSoutullo (changes requested):
Some setters are calling `_reconcile()` and others don't means the builder has two categories of setters with different semantics. That should be explicit, or it'll be easy to misuse later. 
I still think this should be improved and easier to understand. I would rethink about this, and we can split this PR in 3:

1. New Logoscore deployment assets
2. New builders for Logoscore pods/nodes (with tests)
3. New experiment

PR #294, opened 2026-06-13
Add gossip priority queues experiments

mamoutou-diarra:
- Add All priority queues scenarios
- Add methods to generate pod api requester configMap for batch mode
- Allow setting muxer for lsquic benchmark (lsquic can be validated in S0 by setting --muxer to quic)
- Add Analyzer to plot boxplot for delivery delays (group by slow and normal peers)
- Add methods to shape bandwidth using TC in init_container
- Add dual publisher deployments in S4-2: necessary in order inject continuous background traffic and bursty traffic
- improved api_requester builder


src/deployments/experiments/libp2p/gossipsub_priority_queues.py
```diff
@@ -0,0 +1,713 @@
...
+            .with_namespace(self.namespace)
+            .with_cluster_ip("None")
+            .with_publish_not_ready_addresses(True)
+            .with_selector("app", "zerotenkay")
+            .with_port(V1ServicePort(name="p2p-quic", port=5000, target_port=5000, protocol="UDP"))
+            .with_port(V1ServicePort(name="p2p-tcp", port=5000, target_port=5000, protocol="TCP"))
+            .with_port(V1ServicePort(name="metrics", port=8008, target_port=8008, protocol="TCP"))
+            .with_port(V1ServicePort(name="publish", port=8645, target_port=8645, protocol="TCP"))
+            .build()
+        )
```
AlbertoSoutullo: Probably this was pasted here unintentionally?

src/deployments/builders/__init__.py
```diff
@@ -0,0 +1,3 @@
+def __init__(self, **data):
+    super().__init__(**data)
+    self._restart_policy = None
```
AlbertoSoutullo: Why is this here?

src/deployments/libp2p/builders/builders.py
```diff
@@ -24,6 +25,18 @@ class Option:
...
+    gossipsub_d_low = "GOSSIPSUB_D_LOW"
+    gossipsub_d_high = "GOSSIPSUB_D_HIGH"
+    gossipsub_d_out = "GOSSIPSUB_D_OUT"
+    gossipsub_d_lazy = "GOSSIPSUB_D_LAZY"
+    max_high_priority_queue = "GOSSIPSUB_MAX_HIGH_PRIORITY_QUEUE_LEN"
+    max_medium_priority_queue = "GOSSIPSUB_MAX_MEDIUM_PRIORITY_QUEUE_LEN"
+    max_low_priority_queue = "GOSSIPSUB_MAX_LOW_PRIORITY_QUEUE_LEN"
+    slow_peer_penalty_weight = "GOSSIPSUB_SLOW_PEER_PENALTY_WEIGHT"
+    slow_peer_penalty_decay = "GOSSIPSUB_SLOW_PEER_PENALTY_DECAY"
+
```
AlbertoSoutullo: Hmm I don't really agree we should start adding every env variable in here for every experiment. It would be nice to split this somehow imo.

AlbertoSoutullo (changes requested):
Good overall. Next time it would be nice that we split this PR. We could have done new utils/changes + experiments in 2 different PRs. Easier to review.

src/deployments/builders/__init__.py
mamoutou-diarra: should be removed, it should be in the init of the class, not here

src/deployments/libp2p/builders/builders.py
mamoutou-diarra: Ok, I'll keep them only for the  experiment

src/deployments/experiments/libp2p/gossipsub_priority_queues.py
mamoutou-diarra: It was intentional, I took inspiration from connmanager.py and kad_dht.py Is there another way to do it?

AlbertoSoutullo: But, I don't understand why is this return here.
`_deploy_dual_publishers` should only deploy.
You are calling it with await on another place and you are not getting the return value.

PearsonWhite (approved):
Needs rebasing, but conflicts are simple. If you just accept current it should be good.

I didn't do the most thorough review of every single line, but the framework usage is correct and the approach was straightforward 👍

AlbertoSoutullo (approved):

PR #297, opened 2026-06-17
Use pod-api-requester as the publisher for Shadow

radiken:
Replace the traffic_sync.py publisher host with the pod-api-requester run in batch mode.

Related to: https://github.com/vacp2p/10ksim/issues/220


AlbertoSoutullo (changes requested):
I think it would be nice to add some tests to shadow builders (;

AlbertoSoutullo (approved):
Thanks!

PR #298, opened 2026-06-17
Use kad-dht discovery for nimlibp2p regression mesh

radiken:
Add an opt-in kad-dht discovery path to the nimlibp2p experiment. A bootstrap anchor node, headless service, and nodes that discover each other through KAD-DHT instead of the static CONNECTTO dial. Gated by a new `discovery` field.

### DIFF
--- a/src/deployments/experiments/libp2p/nimlibp2p.py
+++ b/src/deployments/experiments/libp2p/nimlibp2p.py
@@ -29,7 +29,7 @@ BOOTSTRAP_NAME = "bootstrap"
 
 
 class ExpConfig(BaseModel):
-    num_nodes: NonNegativeInt = 30
+    num_relay_nodes: NonNegativeInt = 30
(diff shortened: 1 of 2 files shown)


src/deployments/experiments/libp2p/nimlibp2p.py
PearsonWhite: Consider changing this to avoid confusion that `num_nodes` might be total nodes.
```
    num_relay_nodes: NonNegativeInt = 30
```
Assuming the bootstrap nodes are not used as relay nodes (which they aren't).

(Not necessary for merging. Feel free to resolve yourself.)

PearsonWhite (approved):

AlbertoSoutullo (approved):

src/deployments/experiments/libp2p/nimlibp2p.py
radiken: Agree, thank you! Updated.

PR #299, opened 2026-06-18
Add tests for src/deployments/core/builders.py

mamoutou-diarra:
- Added new tests for classes and methods in [builders.py](https://github.com/vacp2p/10ksim/blob/mamoutou/add_tests_deployments_core/src/deployments/core/builders.py).
These tests verify that builder methods correctly set and propagate object configuration. They intentionally do not validate logic implemented in the [configs](https://github.com/vacp2p/10ksim/blob/mamoutou/add_tests_deployments_core/src/deployments/core/configs) package (e.g., build/validation behavior), which should be tested independently.

- Fixed 1 issue in [builders.py](https://github.com/vacp2p/10ksim/blob/mamoutou/add_tests_deployments_core/src/deployments/core/builders.py):
  - `volume_claim_templates` belongs to `stateful_set_spec` not the top-level `StatefulSetConfig`. with_volume_claim_template now initializes `stateful_set_spec.volume_claim_templates` when missing and appends the pvc to the right place.

### DIFF
--- a/src/deployments/core/builders.py
+++ b/src/deployments/core/builders.py
@@ -42,7 +42,7 @@ class StatefulSetBuilder(BaseModel):
         return self
 
     def with_replicas(self, replicas: int) -> Self:
-        self.config.replicas = replicas
+        self.config.stateful_set_spec.replicas = replicas
         return self
 
     def with_label(self, key: str, value: str) -> Self:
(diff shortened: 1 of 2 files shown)


AlbertoSoutullo (approved):
LGTM.

PR #301, opened 2026-06-25
logos core builder changes

PearsonWhite:
Requirements:
1. Allow any order of `.with_` in the caller
2. Allow changing specific fields (such as `.with_service_name`), not just features
3. Incrementally build the config, avoiding giant if/else build() function
4. Allow user to call `.with_` multiple times with the same or different parameters. build() should use the last-called `.with_` values.

The design:
1. Each feature lists its dependencies. See [_apply_logoscore](https://github.com/vacp2p/10ksim/commit/5bf979fb075ded842c9cbde3d3fe2bded3dda1bf#diff-32e95345a8a2b626bdab46ec25c8ce50bd0df8ae535ed542d8ccbf7f42bace02R4-R147)
2. Each feature is applied any time a dependency is changed, but only if all dependencies are set. See [def _reconcile](https://github.com/vacp2p/10ksim/commit/5bf979fb075ded842c9cbde3d3fe2bded3dda1bf#diff-74213642a5f947f6046d61a90ee2424bb113d9b1b10071fbc19a678f0ae0220aR182) in `src/deployments/core/builders.py`
3. When a feature is applied, it is responsible for cleaning up anything it had generated before (eg. dns_config), without removing other changes (eg. explicitly added dns search, or dns search added by another feature). In order to do this, the feature keeps a copy of the previously used parameters. For settings that are overwritten (scalers), they are simply modified. For other settings (lists, etc.), the old value must be found and removed before the new value is added.
4. Each change to a field calls `_reconcile` (in the setters). It's possible we could automatically do this using a decorator for the class.
5. Instead of duplicating certain fields, like `_app`/`app`, I changed the fields into properties and put them into the base class. This is unrelated to the design change.
It's possible we could get the Requirements listed with a better design, but we can't avoid adding significant complexity.

### DIFF
--- a/src/deployments/logos_core/builders/request_builder.py
+++ b/src/deployments/logos_core/builders/request_builder.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from copy import deepcopy
 from dataclasses import dataclass
 from typing import Any, Dict, Optional, Self
 
@@ -17,10 +18,11 @@ from pydantic import Field, PrivateAttr
 
 from src.deployments.core.configs.helpers.identity import apply_identity
 from src.deployments.core.configs.helpers.utils import find_container_config
+from src.deployments.core.dependency_decorator import depends_on
+from src.deployments.experiments.base_experiment import V1Deployable
 from src.deployments.pod_api_requester.builder import (
     PodApiRequesterBuilder,
 )
-from src.deployments.core.dependency_decorator import depends_on
 
 
 @dataclass
@@ -204,10 +206,10 @@ class LogoscorePodApiRequester(PodApiRequesterBuilder):
 
         return self
 
-    def _get_dependencies(self):
-        base_dict = super()._get_dependencies()
-        base_dict["logoscore"] = self._logoscore_dependencies
-        return base_dict
+    def build_dependencies(self) -> Dict[str, V1Deployable]:
+        deps = super().build_dependencies()
+        deps["logoscore"] = self._logoscore_dependencies
+        return deepcopy(deps)
 
 
 def role(namespace: str, role_name: str):
--- a/src/deployments/logos_core/builders/request_builder.py
+++ b/src/deployments/logos_core/builders/request_builder.py
@@ -131,7 +131,6 @@ class LogoscorePodApiRequester(PodApiRequesterBuilder):
 
     def with_debug(self, debug: bool = True) -> Self:
         self.debug = debug
-        self._reconcile("debug")
         return self
 
     @depends_on(
--- a/src/deployments/logos_core/builders/request_builder.py
+++ b/src/deployments/logos_core/builders/request_builder.py
@@ -102,7 +102,7 @@ class LogoscorePodApiRequester(PodApiRequesterBuilder):
     def logoscore_enabled(self) -> bool:
         return self._logoscore_enabled
 
-    @debug.setter
+    @logoscore_enabled.setter
     def logoscore_enabled(self, logoscore_enabled: bool) -> None:
         self._logoscore_enabled = logoscore_enabled
         self._reconcile("logoscore_enabled")
--- a/src/deployments/logos_core/builders/request_builder.py
+++ b/src/deployments/logos_core/builders/request_builder.py
@@ -20,9 +20,7 @@ from src.deployments.core.configs.helpers.identity import apply_identity
 from src.deployments.core.configs.helpers.utils import find_container_config
 from src.deployments.core.dependency_decorator import depends_on
 from src.deployments.experiments.base_experiment import V1Deployable
-from src.deployments.pod_api_requester.builder import (
-    PodApiRequesterBuilder,
-)
+from src.deployments.pod_api_requester.builder import PodApiRequesterBuilder
 
 
 @dataclass
@@ -170,7 +168,6 @@ class LogoscorePodApiRequester(PodApiRequesterBuilder):
         if old_params and old_params.service_name and old_params.namespace:
             old_search = f"{old_params.service_name}.{old_params.namespace}.svc.cluster.local"
             pod_spec.remove_dns_search(old_search, missing_ok=True)
-
         new_search = f"{new_params.service_name}.{new_params.namespace}.svc.cluster.local"
         pod_spec.with_dns_search(new_search, overwrite=True)
 
--- a/src/deployments/logos_core/builders/request_builder.py
+++ b/src/deployments/logos_core/builders/request_builder.py
@@ -1,184 +1,213 @@
-from collections import defaultdict
-from copy import deepcopy
-from itertools import chain
-from typing import Any, Dict, Optional, Self, Set
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Dict, Optional, Self
 
 from kubernetes.client import (
     RbacV1Subject,
-    V1EnvVar,
     V1ObjectMeta,
     V1PodSecurityContext,
     V1PolicyRule,
-    V1ResourceRequirements,
     V1Role,
     V1RoleBinding,
     V1RoleRef,
-    V1Service,
     V1ServiceAccount,
-    V1ServicePort,
-    V1ServiceSpec,
 )
 from pydantic import Field, PrivateAttr
 
-from src.deployments.core.configs.container import ContainerConfig
 from src.deployments.core.configs.helpers.identity import apply_identity
-from src.deployments.core.configs.helpers.utils import find_container_config, get_config
-from src.deployments.core.configs.pod import PodSpecConfig
-from src.deployments.core.k8s_object import V1Deployable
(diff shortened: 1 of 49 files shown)


src/deployments/logos_core/builders/request_builder.py
```diff
@@ -1,184 +1,213 @@
...
+    @debug.setter
+    def debug(self, value: bool) -> None:
+        self._debug = value
+        self._reconcile("debug")
+
+    @property
+    def logoscore_enabled(self) -> bool:
+        return self._logoscore_enabled
+
+    @debug.setter
```
AlbertoSoutullo: Shouldn't this be `logoscore_enabled.setter`?

src/deployments/logos_core/builders/request_builder.py
```diff
@@ -1,184 +1,213 @@
...
-def service_account(namespace: str, service_account_name: str):
-    return V1ServiceAccount(
-        api_version="v1",
-        kind="ServiceAccount",
-        metadata=V1ObjectMeta(name=service_account_name, namespace=namespace),
-    )
+    def _get_dependencies(self):
+        base_dict = super()._get_dependencies()
+        base_dict["logoscore"] = self._logoscore_dependencies
+        return base_dict
```
AlbertoSoutullo: If the rest of the dependency model expects values like lists of deployables by kind (services, roles, role_bindings, etc.), then why we have a "logoscore" here? Just to understand.

src/deployments/pod_api_requester/builder.py
```diff
@@ -15,45 +18,184 @@
...
+            old_params=self._base_requester_params, new_params=new_params
+        )
+        self._base_requester_params = new_params
+        return self
+
+    def _apply_requester_base_inner(
+        self, old_params: RequesterBaseParams, new_params: RequesterBaseParams
+    ):
+        config = self.config
+
```
AlbertoSoutullo, line 139: Changing `service_name` won't trigger re-application right? Its not in the depencendy list but `_apply_requester_base_inner` uses `searches=[f"{new_params.service_name}.{new_params.namespace}.svc.cluster.local"]`

AlbertoSoutullo (changes requested):
Looks much better tbh.

Tests are missing.

src/deployments/logos_core/builders/request_builder.py
PearsonWhite: I decided to put the logoscore specific dependencies in their own sub-dict just to organize them by feature. This also makes them independent of how the base dependencies are organized. For example, if the base dependencies had:
```
{
  "roles" : single_role,
  "role_bindings" : [role_binding_1, role_binding_2],
}
```
Not necessarily a super realistic concern, but I figured it doesn't hurt to just put them in their own sub-dict.

src/deployments/pod_api_requester/builder.py
PearsonWhite: Fixed in 9ea904094f289366ea0eefc5c38a392eb314e85d

src/deployments/logos_core/builders/request_builder.py
PearsonWhite: Fixed in 9ea904094f289366ea0eefc5c38a392eb314e85d

src/deployments/logos_core/tests/test_experiment.py
```diff
@@ -0,0 +1,61 @@
...
+    experiment = LogosDeliveryExperiment.model_construct(
+        config=exp_config,
+        api_client=Mock(),
+        namespace="ns",
+    )
+
+    actual = collect_actual_resources(experiment)
+    expected = load_expected_case(case_dir)
+
+    assert expected == actual
```
AlbertoSoutullo, line 61: In `src/deployments/logos_core/tests/test_experiment.py`, both helper functions are annotated/structured as if they return dicts, but they return sorted(result), which returns a list of keys, not the dict contents.

That means this test likely compares only filenames, not rendered YAML objects.

So this:
```
load_expected_case(...) returns sorted keys
collect_actual_resources(...) returns sorted keys
assert expected == actual
```
can pass even if the actual generated resource content is wrong, as long as the filenames match.

AlbertoSoutullo (changes requested):
What’s still missing are direct tests for the actual requirements, ie:

- calling .with_* in different orders yields same final result
- repeated .with_* calls use the last value
- changing a field re-applies the relevant feature and removes prior generated state correctly
- changing one feature’s generated list entries doesn’t accidentally remove user-added or other-feature-added entries

For example, would be nice to have explicit tests like:

- call `.with_service_name("a")`, then `.with_service_name("b")`, verify only b DNS entry remains
- call `.with_logoscore()` before namespace/name/app/service_name, then fill them in later, verify final build matches the normal order
- call `.with_dns_search(user_value)` before and after feature application, verify feature cleanup does not remove user-managed entries

etc.

src/deployments/logos_core/tests/test_experiment.py
PearsonWhite: Fixed in aa12aecaba77d939bfe7aea05dc59f9cbe6c684b

src/deployments/logos_core/builders/request_builder.py
```diff
@@ -1,184 +1,215 @@
...
+        self.pod_service_reader_binding_name = binding_name
         return self
 
-    def with_service_name(self, service_name: str) -> Self:
-        self._service_name = service_name
-        self._reconcile()
+    def with_debug(self, debug: bool = True) -> Self:
+        self.debug = debug
+        self._reconcile("debug")
         return self
```
AlbertoSoutullo, line 132: Since the property setter already calls _reconcile("debug"), this is redundant.

AlbertoSoutullo (changes requested):
The current tests demonstrate outputs for a couple of happy paths. The tests I requested in the previous comment are still not there. 
I’m not asking for exhaustive coverage of every permutation, but just 3–4 focused tests that lock down the intended behaviors.

src/deployments/logos_core/builders/request_builder.py
PearsonWhite: Fixed in c7751ee49f4519e47757808e942696913c0230ee

src/deployments/logos_core/tests/test_request_builder.py
```diff
@@ -0,0 +1,219 @@
...
+def test_order_independences(func_list):
+    builder = LogoscorePodApiRequester()
+    for fun in func_list:
+        fun(builder)
+
+    check_deployments_outputs("base_case", builder)
+
+
+def check_dns_config(deployment: V1Pod, expected_dns_config: List[str]):
+    deployment.spec.dns_config == expected_dns_config
```
AlbertoSoutullo: This function doesn't do anything.

src/deployments/logos_core/tests/test_request_builder.py
```diff
@@ -0,0 +1,219 @@
...
+    def test_multiple_calls(self, base_builder):
+        """Test that we can override default service account name."""
+        pod = (
+            base_builder.with_app("app")
+            .with_logoscore()
+            .with_service_account_name("custom-service-account")
+            .with_service_account_name("custom-service-account2")
+            .build()
+        )
+        pod.spec.service_account_name == "custom-service-account2"
```
AlbertoSoutullo: These functions are not asserting anything

AlbertoSoutullo: Also, `test_after_feature` is defined twice

AlbertoSoutullo (changes requested):

src/deployments/logos_core/tests/test_request_builder.py
PearsonWhite: Removed in 773fb771a1f652cc89732ca0e08446937f49d00c

src/deployments/logos_core/tests/test_request_builder.py
PearsonWhite: > Also, `test_after_feature` is defined twice

773fb771a1f652cc89732ca0e08446937f49d00c

> These functions are not asserting anything

We're asserting the behavior of calling `with_service_account_name` multiple times and with `with_logoscore`. `service_account_name` is added by `with_logoscore`, but we should be able to override the default. We can do this before calling `with_logoscore` or after. We can also set it multiple times and the most recently called `with_service_account_name` should be used.

AlbertoSoutullo: What I meant is we are literally not asserting, assert is not there.
We have:
`pod.spec.service_account_name == "custom-service-account"`

I would assume we want:
`assert pod.spec.service_account_name == "custom-service-account"`

PearsonWhite: I don't know how I missed that 🤦‍♂️ Fixed in b8ccdfe138bfeac7870e9b2ba7fd3c09008b6d65

AlbertoSoutullo (approved):

PR #303, opened 2026-06-25
Update str with Paths

AlbertoSoutullo:
Minor change for using Paths. In next PR will spleat Bridge.py file and add tests.


radiken (approved):
LGTM

PR #304, opened 2026-06-25
BaseExperiment deploy: allow multiple deployments at once

PearsonWhite:
This lets us deploy the dependencies and pod/statefulset/etc. all at once
```
bootstrap_deployments = self.build_node_deployments("bootstrap")
await self.deploy(bootstrap_deployments, exist_ok=True)
```

### DIFF
--- a/src/deployments/experiments/tests/test_base_experiment.py
+++ b/src/deployments/experiments/tests/test_base_experiment.py
@@ -17,6 +17,12 @@ class DummyExperiment(BaseExperiment[DummyCfg]):
         pass
 
 
+class DeploymentObj:
+    def __init__(self, kind, name):
+        self.kind = kind
+        self.metadata = {"name": name, "namespace": "ns"}
+
+
 @pytest.mark.asyncio
 @pytest.mark.parametrize("strategy", ["serial", "parallel"])
 async def test_deploy_orders_base_resources_before_workloads(strategy):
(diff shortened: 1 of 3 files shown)


src/deployments/experiments/tests/test_base_experiment.py
```diff
@@ -0,0 +1,123 @@
...
+
+    items = [
+        DeploymentObj("Pod", "pod1"),
+        DeploymentObj("RoleBinding", "rb1"),
+        DeploymentObj("StatefulSet", "sts1"),
+        DeploymentObj("ServiceAccount", "sa1"),
+        DeploymentObj("Service", "s"),
+        DeploymentObj("DaemonSet,", "ds"),
+        DeploymentObj("Job,", "j"),
+        DeploymentObj("CronJob,", "cj"),
```
radiken:
```suggestion
        DeploymentObj("DaemonSet", "ds"),
        DeploymentObj("Job", "j"),
        DeploymentObj("CronJob", "cj"),
```

radiken (approved):
LGTM, just one minor comment

AlbertoSoutullo (changes requested):
Good improvement. Minor change.

src/deployments/utils/flatten.py
```diff
@@ -0,0 +1,9 @@
+def flatten(obj):
+    if isinstance(obj, dict):
+        for value in obj.values():
+            yield from flatten(value)
+    elif isinstance(obj, list):
+        for value in obj:
+            yield from flatten(value)
+    else:
+        yield obj
```
AlbertoSoutullo, line 9: This function handles `dict` and `list` but the new type hint says `Collection[V1Deployable]`, and `Collection` is broader.
So we either wide flatten to support other shapes or narrow the type hint.

AlbertoSoutullo (changes requested):
Forgot to submit the review 🤦

src/deployments/utils/flatten.py
AlbertoSoutullo, line 9: This function handles `dict` and `list` but the new type hint says `Collection[V1Deployable]`, and `Collection` is broader.
So we either wide flatten to support other shapes or narrow the type hint.

AlbertoSoutullo (changes requested):

src/deployments/utils/flatten.py
PearsonWhite: Changed in 1145d4c3e5f15d2207ae2c9610753a2e19ab67bc

AlbertoSoutullo: I didn’t mean to make this recurse into any Iterable, I think that’s too broad. My intent was only to support container inputs like dict, list, tuple, and maybe set. I would restrict `flatten()` to those explicit container types, or I would also narrow the deploy() type hint accordingly if that makes sense.

AlbertoSoutullo (changes requested):

src/deployments/utils/flatten.py
PearsonWhite: Replaced with this commit: 240c7c4cda44ad03a2f563cf7125b7401072de78

AlbertoSoutullo (approved):

PR #306, opened 2026-06-26
Split base bridge and tests

AlbertoSoutullo:
I am planning to extend base bridge so before doing it, I am cleaning and adding tests to it.


radiken (approved):
LGTM!

PR #307, opened 2026-06-26
Window event

AlbertoSoutullo:
The rationale is to factor out a repeated bridge pattern: “given an events log, define named time windows from event markers, optionally shift the bounds, then expose the selected interval as stack start/end.”

Before this abstraction, each protocol bridge tends to repeat the same shape:
```python
complete.start = wait_for_clear_finished
complete.end = internal_run_finished + 30s
stable.start = start_messages + 3m
stable.end = publisher_messages_finished - 30s
```

`EventWindowEndpoint` is the small data model for one bound of a window:
```
key: dict
time_shift: timedelta
```

So instead of manually constructing `EventMapping(...)` everywhere, callers can write:
```
"stable": {
    "start": event_window("start_messages", timedelta(minutes=3)),
    "end": event_window("publisher_messages_finished", timedelta(seconds=-30)),
}
```

`EventWindowBridge` is the reusable bridge base for classes where metadata is driven by those windows. It does three things:
1. Converts event_windows into EventMappings.
2. Calls the existing _get_metadata_from_events_list(...) machinery.
3. Copies the selected interval into:
```
metadata["stack"]["start_time"]
metadata["stack"]["end_time"]
metadata["results"]
metadata["stack"]["container_name"]
```

That lets bridges for waku, libp2p or service discovery become declarative, they only define container_name, available windows, and default selected interval.

### DIFF
--- a/src/deployments/libp2p/bridge.py
+++ b/src/deployments/libp2p/bridge.py
@@ -1,46 +1,37 @@
-# Python imports
+# Python Imports
 import logging
 from datetime import timedelta
-from pathlib import Path
-from typing import Literal
+from typing import Dict, Literal
 
 # Project Imports
-from src.deployments.core.base_bridge import BaseBridge, EventMapping
+from src.deployments.core.event_window_bridge import (
+    EventWindowBridge,
+    EventWindowEndpoint,
+    event_window,
+)
 from src.deployments.libp2p.builders.helpers import LIBP2P_CONTAINER_NAME
 
 logger = logging.getLogger(__name__)
 
 
-class Bridge(BaseBridge):
+class Bridge(EventWindowBridge):
     interval: Literal["complete", "stable"] = "complete"
     """Time interval for start and end times.
 
     complete: From the beginning to the end of the whole experiment run code
 
     stable: Just the time where nodes should have stablized.
-    Calculated in _get_metadata_events by adding an offset from when message publishing starts and ends.
+    Calculated by adding offsets to message publishing start/end events.
     """
 
-    def get_metadata(self, events_log_path: str) -> dict:
-        metadata = super().get_metadata(events_log_path)
-        events = self._get_metadata_events(events_log_path)
-        metadata["results"] = events
-        metadata["stack"]["start_time"] = events[self.interval]["start"]
-        metadata["stack"]["end_time"] = events[self.interval]["end"]
-        metadata["stack"]["container_name"] = LIBP2P_CONTAINER_NAME
-        return metadata
-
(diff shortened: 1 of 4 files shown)


src/deployments/libp2p/bridge.py
```diff
@@ -1,46 +1,37 @@
...
-        ]
-        events_list = list(
-            map(
-                lambda obj: EventMapping(key={"event": obj[0]}, target=obj[1], time_shift=obj[2]),
-                events,
-            )
-        )
-        return self._get_metadata_from_events_list(events_log_path, events_list)
+    container_name: str = LIBP2P_CONTAINER_NAME
+    event_windows: Dict[str, Dict[str, EventWindowEndpoint]] = {
```
PearsonWhite: Calling `event_window` for what is a bound seems wrong. I think it would make more sense if we ended up with something like this:
Name EventWindowEndpoint -> EventBound
```
    event_windows: Dict[str, Dict[str, EventBound]] = {
        "complete": {
            EventBound("start": "wait_for_clear_finished"),
            EventBound("end": "internal_run_finished", timedelta(seconds=30)),
        },
        "stable": {
            EventBound("start": "start_messages", timedelta(minutes=3)),
            EventBound("end": "publisher_messages_finished", timedelta(seconds=-30)),
        },
    }
```
But do we need a translation layer here? Can we end up with something more like this?
```
    event_windows: List[EventWindow] = [
        EventWindow(key="complete",
            start=EventBound("wait_for_clear_finished"),
            end=EventBound("internal_run_finished", timedelta(seconds=30)),
        ),
        EventWindow("stable",
            start=EventBound("start_messages", timedelta(minutes=3)),
            end=EventBound("publisher_messages_finished", timedelta(seconds=-30)),
        ),
    ]
```

PearsonWhite (changes requested):
I don't see why event_windows needs to be in a field instead of simply defining it in the function. This adds a layer of inheritance that probably doesn't need to be there.

src/deployments/libp2p/bridge.py
AlbertoSoutullo: Yeah, I see your point.
[9f40f61](https://github.com/vacp2p/10ksim/pull/307/commits/9f40f618bb73013151436e4da63d09bf05ca6c4c)

PearsonWhite: Changes look good here.

PearsonWhite (approved):
Looks good. It's fine to merge, but consider https://github.com/vacp2p/10ksim/pull/307#pullrequestreview-4619128365

PR #308, opened 2026-06-26
Post run analysis

AlbertoSoutullo:
This PR defines the post-run analysis flow as an import-string based hook on `BaseExperiment`.

Experiments can set automatic post-run analysis by setting:

```python
post_run_analysis = "module.path:function_name"
```

BaseExperiment.run() now runs that analysis after final metadata is generated and written. The analysis receives the finalized experiment instance, including experiment.metadata, metadata_log_path, and output_folder.


PearsonWhite (commented):
I don't think this should be a `_run` function. Thus far, we've kept the deployment, event-reading/metadata creation/bridge, and analyzer logic in separate objects, which I think is good. The experiment shouldn't be responsible for the analysis logic.

Since we have `_type` information in the `BaseExperiment`, we can, when analyzing, serialize the experiment (like here: https://github.com/vacp2p/10ksim/blob/a8c8fa667b22954a36e8545e275c85295d5df5b2/full_run.py#L107) and use the `_type` metadata to fetch the appropriate analyzer.

That way, we can choose to have the Experiment -> Analyzer mapping as either a separate table, or simply a field in the Experiment. Either way, we could easily add a CLI interface to something like (https://github.com/vacp2p/10ksim/blob/pwhite/unified/full_run.py) and allow overriding the analyzer_type+config the same way we do for `ExpConfig` in `BaseExperiment[ExpConfig]`.

radiken (changes requested):
`_run_analysis` exists on two experiments, so I guess we should migrate them in this PR as well. 
- `connmanager` defines it as `_run_analysis(self, config, start_time, end_time)` and calls it in `_run()`, so the new no-arg call in `BaseExperiment.run()` will fail.
- `shadow_gossipsub` overrides `run()` and calls `_run_analysis()` itself, so it will run twice.

src/deployments/experiments/base_experiment.py
```diff
@@ -347,6 +349,7 @@ async def run(self):
 
         self.log_event("run_finished")
         self._dump_metadata()
+        run_post_analysis(self)
```
PearsonWhite: This will block continued deployment runs if we call a multi-experiment. Can we avoid waiting on all the data pulling to run the next experiment?

PearsonWhite (changes requested):
Instead of using a str, loading the module, and getting the function, wouldn't it be easier to either:
1. Have an instance method in the experiment. This could be in the form of another class, like we do for metadata gathering:
```
    def _get_metadata(self) -> dict:
        return Bridge().get_metadata(self.events_log_path)
```
2. Directly set the class and de/serialize like we do for BaseExperiment._type if we want to serialize which analyzer we used.

The only advantage I can see for using a str here is making de/serialization easier, but then we have a this chunk of logic to convert it to the actual function.

src/deployments/experiments/base_experiment.py
AlbertoSoutullo: Good point.

src/analysis/post_run/connmanager.py
```diff
@@ -0,0 +1,73 @@
...
+            "type": "vaclab",
+            "url": VICTORIA_LOGS_URL,
+            "reader": "victoria",
+            "stateful_sets": ["hub"],
+            "nodes_per_statefulset": [1],
+            "container_name": "pod-0",
+            "namespace": experiment.namespace or stack.get("namespace"),
+            "extra_fields": ["kubernetes.pod_name"],
+        }
+    )
```
radiken, line 52: This stack has no `start_time` and `end_time`, so `VictoriaQueryBuilder` drops the `_time` filter. The old `_run_analysis` bounded the query to the run window, without it the fixed `nimlibp2p` and `pod-0` selectors match every past run.

Could we reconstruct the window in the analyzer (`run-start` and `finish` are already in the metadata), or have the bridge carry `start_time` and `end_time` into `metadata["stack"]`?

radiken (changes requested):

src/analysis/post_run/connmanager.py
AlbertoSoutullo: So, in a normal run connmanager should get `start_time` and `end_time` from `Libp2pBridge` because `run_connmanager_analysis()` consumes `experiment.metadata["stack"]`. So I don’t think the analyzer should reconstruct the window.

It is true that if we only want to re-run the analysis, this doesn't exist. I’ll add validation so connmanager analysis refuses to run if the metadata stack is missing `start_time` or `end_time` to avoid Victoria doing things that we don't want to.

For the re-analysis only workflow should be a separate CLI entry point that reads `metadata.json`, reconstructs the experiment with `read_experiment(...)`, and calls `run_post_analysis(experiment)`. That lets us re-run analysis without re-running deployment, while still using the finalized metadata window. That will be done in a separate PR.

radiken (approved):
LGTM!

PearsonWhite (approved):

PR #310, opened 2026-06-28
Improve experiment config CLI params

### LINKED ISSUE #153: General tooling (recurring)

PearsonWhite:
* Automatically generate params for experiment configs
* Easier to extend BaseExperiment subparser

Before:
```
 python ./deployment.py waku --help
usage: deployment.py waku [-h] [--skip-check] [--dry-run] [--namespace NAMESPACE]

options:
  -h, --help            show this help message and exit
  --skip-check          If present, does not wait until the namespace is empty before running the test.
  --dry-run             If True, does not actually deploy kubernetes configs but run kubectl apply --dry-run.
  --namespace NAMESPACE
                        The namespace for deployments.
```

After:
```
> python ./deployment.py waku --help
usage: deployment.py waku [-h] [--skip-check] [--dry-run] [--namespace (str)] [--num-messages (int)] [--num-nodes (int)] [--num-bootstrap-nodes (int)] [--delay-cold-start (float)]
                          [--delay-after-publish (float)]

options:
  -h, --help            show this help message and exit
  --skip-check          If present, does not wait until the namespace is empty before running the test.
  --dry-run             If True, does not actually deploy kubernetes configs but run kubectl apply --dry-run.
  --namespace (str)     The namespace for deployments.
  --num-messages (int)
  --num-nodes (int)
  --num-bootstrap-nodes (int)
  --delay-cold-start (float)
  --delay-after-publish (float)
```
Now we can do this:
```python ./deployment.py nimlibp2p --num-relay-nodes=80 --namespace zerotesting```
Without having to explicitly add the `--num-relay-nodes` param. It is automatically derived from the field in ExpConfig:
```
class ExpConfig(BaseModel):
    model_config = ConfigDict(use_attribute_docstrings=True)

    num_relay_nodes: NonNegativeInt = 30
```

### DIFF
--- a/src/deployments/experiments/tests/base_experiment/test_cli_parser.py
+++ b/src/deployments/experiments/tests/base_experiment/test_cli_parser.py
@@ -253,3 +253,21 @@ def _test_help_string(
 
     for expected_string in expected:
         assert expected_string in captured.out
+
+
+def test_cannot_override_params():
+    parser = argparse.ArgumentParser(description="Test parser", formatter_class=_HelpFormatter)
+    subparsers = parser.add_subparsers(dest="experiment", required=True)
+    with pytest.raises(TypeError):
+
+        class NoOverrideParamExp(BaseExperiment[EmptyExpConfig]):
+            model_config = ConfigDict(use_attribute_docstrings=True)
+
+            name: ClassVar[str] = "NoOverrideParamExp"
+
+            @classmethod
+            def add_parser(cls, subparsers) -> None:
+                subparser = subparsers.add_parser(cls.name, help=cls.__doc__)
+
+        NoOverrideParamExp.add_parser(subparsers)
+        _exp = NoOverrideParamExp()
--- a/src/deployments/experiments/tests/base_experiment/test_cli_parser.py
+++ b/src/deployments/experiments/tests/base_experiment/test_cli_parser.py
@@ -1,6 +1,6 @@
 import argparse
 import logging
-from typing import ClassVar, List, Literal, Optional, Type
+from typing import ClassVar, List, Literal, Type
 
 import pytest
 from pydantic import BaseModel, ConfigDict, Field, NonNegativeFloat, NonNegativeInt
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: src/deployments/utils/parser.py [hunk 1]
```diff
@@ -0,0 +1,82 @@
...
+    origin = get_origin(annotation)
+    if origin is Union:
+        args = [arg for arg in get_args(annotation) if arg is not type(None)]
+        if len(args) == 1 and args[0] in (int, float, str):
+            kwargs["type"] = args[0]
+
+    if origin is Literal:
+        choices = get_args(annotation)
+        if choices:
+            kwargs["type"] = type(choices[0])
```
AlbertoSoutullo, line 63: I mean with this behavior argparse help suggests constrained values, but argparse itself won’t reject invalid ones at parse time, because it's just  setting `type` and not `choices`.

src/deployments/utils/parser.py
```diff
@@ -0,0 +1,82 @@
+# Python Imports
+from typing import Any, Literal, Union, get_args, get_origin
+
+from pydantic import BaseModel
+from pydantic.fields import FieldInfo
+
+ARG_NOT_SET = object()
+
+
+from typing import Any, Literal, Union, get_args, get_origin
```
AlbertoSoutullo: Duplicated import

src/deployments/experiments/tests/base_experiment/test_cli_parser.py
```diff
@@ -0,0 +1,291 @@
...
+        )
+        cls.add_args(subparser)
+        subparser.add_argument(
+            "--custom-arg",
+            type=str,
+            required=False,
+            default="custom-arg",
+            metavar="(str)",
+            help="A custom arg added via a custom add_parser function",
+        )
```
AlbertoSoutullo: Related to the main comment, test says "It is acceptable for a subclass to replace add_parser() and manually build a parser without base args or config args."

I would rewrite the tests so they validate the propper intention.
For a normal subclass using the framework:

1. inherited `add_parser()` includes base args
2. inherited `add_parser()` includes config-derived args
3 overriding `add_args()` adds custom args on top

Imo that is the behavior worth checking

AlbertoSoutullo (changes requested):
There is a weird change here. So now you are saying that every experiment should automatically expose CLI params derived from ExpConfig.
But with the current design, a subclass can still do this:
```python
@classmethod
def add_parser(cls, subparsers):
    subparser = subparsers.add_parser(cls.name)
    subparser.add_argument("--my-custom-arg")
```
and silently lose `--skip-check`, `--dry-run`, all auto-generated config args...
Maybe I would make `add_parser()` the stable framework method, and move all customization into hooks.

src/deployments/utils/parser.py
PearsonWhite: Fixed in 161d0644218af1411e4b5ac54e5c979acf48b202

src/deployments/utils/parser.py
PearsonWhite: Fixed in 40963978da9cdb33adc163cffe20459e07f7b4f4

src/deployments/experiments/tests/base_experiment/test_cli_parser.py
PearsonWhite: I added a check to make sure there are no base flags in this commit: 34ccfd605cbe96777fdf7f99f7d21ffaaef7e7ad

- In `test_experiments` we check that the base flags are there for each test.
- For `test_experiment_custom`, we now check that base flags are not there.
- `CustomArgsExperiment` checks that `add_args` adds the args on top of the base.

I'm not sure I fully understand your comment.

AlbertoSoutullo: I think the confusion is that my concern is not about if a custom `add_parser()` can omit the base flags. My point is that I don’t think we should codify that as intended framework behavior in the tests. I’d prefer `add_parser()` to remain the framework-owned method, and customization to happen through hooks like `add_args()`.
Because `add_parser()` is the entry point that assembles the full CLI. If subclasses override that method directly, they can accidentally skip framework behavior.

PearsonWhite: Removed tests for `add_parser` in 1356f5f50f5a109028c739456b5347090369854c and e456b46274f078bb934873a00f5a34010661c9ad.

- Ensure that derived classes do not override `add_parser`
- Remove custom `add_parser`s in existing experiments
- Do not generate duplicate flags

AlbertoSoutullo (approved):

PR #315, opened 2026-07-03
Static mesh formation toggle

radiken:
As discussed with Alberto, "it would be nice to have a mesh formation option in the node, like static or kad.dht". This PR is to enable that.

`discovery` already toggles static vs kad-dht mesh formation, but the static path was only half-wired: `_run` creates a discovery service for kad-dht only, so choosing static leaves nodes unable to resolve `nimp2p-service` and the rollout deadlocks.

- Restore `build_static_service()` (headless `nimp2p-service`, `publish_not_ready_addresses=True` so not-yet-ready pods are resolvable at cold start) and deploy it for static discovery. This service was removed as "unused" in b233dc2; it's needed again now that static is useful for isolating kad-dht vs gossipsub behaviour.
- Add `wait_nodes_ready` (default True): we should be able to rely on readiness probe for both static and discovery with any muxer.

### DIFF
--- a/src/deployments/experiments/libp2p/nimlibp2p.py
+++ b/src/deployments/experiments/libp2p/nimlibp2p.py
@@ -46,6 +46,9 @@ class ExpConfig(BaseModel):
     network_delay: NonNegativeInt = 0
     network_jitter: NonNegativeInt = 0
     node_start_delay: NonNegativeInt = 60
+    wait_nodes_ready: bool = True
+    """Gate publishing on a ready rollout; set False for sparse meshes (static,
+    quic) whose readiness probe flaps and never reaches a stable 100%."""
 
     @model_validator(mode="after")
     def _check_bootstrap_nodes(self):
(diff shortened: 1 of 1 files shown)


AlbertoSoutullo (approved):
LGTM in 10ksim. Is the code for the node ready for this beheavior tho?

src/deployments/experiments/libp2p/nimlibp2p.py
```diff
@@ -46,6 +46,9 @@ class ExpConfig(BaseModel):
     network_delay: NonNegativeInt = 0
     network_jitter: NonNegativeInt = 0
     node_start_delay: NonNegativeInt = 60
+    wait_nodes_ready: bool = True
+    """Gate publishing on a ready rollout; set False for sparse meshes (static,
```
PearsonWhite: Is this based on the muxer then? Should we validate that?
The PR says: `Static/quic runs set it False and rely on delay_cold_start as a fixed formation window.`, but it looks like this is just a manual option.

AlbertoSoutullo: Well I didn't notice this. Nodes in quic should be ready. If we want static connection with quic, what we need is the ping.

radiken: Good catch! Made it derived instead of a manual option. `wait_nodes_ready` is now Optional and defaults off discovery (static → False, kad-dht → True), and you can still pass an explicit value to override.

I put it on discovery rather than muxer because the readiness probe passes only when gossipsub `healthy_peers_topics == 1`, i.e. the test mesh is at or above `dLow`. A static run only dials `connect_to` (10) peers so the mesh sits right around dLow and keeps dipping under, which flaps the probe. Kad-dht discovers enough peers to stay above it. quic just makes the static case worse.

Agree the cleaner fix for static quic is ping, but that's node-side, so will do a follow up there.

AlbertoSoutullo: @radiken But wait, I think I am missing something. Why don't we want the `wait_nodes_ready` off in static? Both in static and with discovery should be true before moving to the next step.

radiken: Did some more tests, you are right. We can actually always keep it on. Forget about what I said above, doesn't make sense from what I tested.

PR #343, opened 2026-07-06
Logoscore: experiment

PearsonWhite:


src/deployments/logos_core/experiment.py
```diff
@@ -0,0 +1,403 @@
...
+            except Exception as e:
+                logger.error(f"e: {e}")
+            await init_node(self.namespace, indexed_name, self.relay_service, addresses)
+
+        for name, service in zip(
+            [bootstrap_name, relay_name], [self.bootstrap_service, self.relay_service]
+        ):
+            for index in range(self.config.num_relay_nodes):
+                indexed_name = f"{name}-{index}"
+                await start_node(self.namespace, indexed_name, service)
```
AlbertoSoutullo: Aren't we using then the same number of relay nodes and bootstrap nodes here?

src/deployments/logos_core/experiment.py
```diff
@@ -0,0 +1,403 @@
...
+        topic = "/my-app/1/dst/proto"
+        for index in range(self.config.num_relay_nodes):
+            indexed_name = f"{relay_name}-{index}"
+            await subscribe(self.namespace, indexed_name, self.relay_service, topic)
+
+        message = "aGVsbG8="  # Test message
+        for _message_index in range(self.config.num_messages):
+            index = random.randrange(0, self.config.num_relay_nodes)
+            indexed_name = f"{relay_name}-{index}"
+            await send(self.namespace, indexed_name, self.relay_service, topic, message)
```
AlbertoSoutullo: We are not using `delay_after_publish` in here, is it intended?

AlbertoSoutullo (changes requested):

src/deployments/logos_core/experiment.py
PearsonWhite: Fixed in d1ef969be54f750a76f6893427afeba6f88de34a + 51782ea5af570bb492d5805e3ab4a8db03979692

src/deployments/logos_core/experiment.py
PearsonWhite: Fixed in d1ef969be54f750a76f6893427afeba6f88de34a

PR #344, opened 2026-07-06
WakuExperiment: updates

PearsonWhite:
This is more of a temporary patch to encapsulate some needed changes for running the latest nWaku versions. I expect this logic to change at some point. Hence, it may seem incomplete.

---

Update WakuExperiment

* Add ExpConfig params for image, etc.
* Add option to grab /metrics
* Update builders for new regression params
* Add MultiWakuExperiment with some known versions

Update WakuTracer to support JSON formatted logs

Sometime after nWaku v0.37.0-rc.4, logs the chronicles logging library
adding color fomratting via ANSI escape sequences resulting in difficult
to read logs. To fix this, add this arg to the nWaku node:
`--log-format=JSON`. This change allows the tracer to read the new log
format.

### DIFF
--- a/src/analysis/mesh_analysis/readers/tracers/tests/test_waku_tracer.py
+++ b/src/analysis/mesh_analysis/readers/tracers/tests/test_waku_tracer.py
@@ -1,0 +1,90 @@
+import re
+
+import pandas as pd
+
+from src.analysis.mesh_analysis.readers.tracers.waku_tracer import WakuTracer
+
+
+def mock_queries(data, tracer):
+    results = [[] for _ in tracer.patterns]
+    for i, pattern_group in enumerate(tracer.patterns):
+        query_results = [[] for _ in pattern_group.trace_pairs]
+        for log_line in data:
+            for j, trace_pair in enumerate(pattern_group.trace_pairs):
+                pattern = trace_pair.regex
+                match = re.search(pattern, log_line[0])
+                if match:
+                    match_as_list = list(match.groups())
+                    match_as_list.extend(log_line[1:])
+                    query_results[j].append(match_as_list)
+            results[i].extend(query_results)
+    return results
+
+
+def test_received_regex_matches_text():
+    tracer = WakuTracer().with_received_pattern_group()
+
+    line = (
+        "received relay message my_peer_id=16U*GiNg1a "
+        "msg_hash=0x17cfd30767acac9b86c18333ba918abef93cc23f65b6c98c845c682584f92583 "
+        "from_peer_id=16U*wJXtuH receivedTime=1763643976019361536"
+    )
+
+    parsed_logs = mock_queries([[line]], tracer)
+    dfs = tracer.trace(parsed_logs)
+    received_df = dfs["received"][0]
+
+    assert list(received_df.columns[:4]) == [
+        "receiver_peer_id",
+        "msg_hash",
+        "sender_peer_id",
+        "timestamp",
+    ]
+    assert received_df.loc[0, "receiver_peer_id"] == "16U*GiNg1a"
+    assert (
+        received_df.loc[0, "msg_hash"]
+        == "0x17cfd30767acac9b86c18333ba918abef93cc23f65b6c98c845c682584f92583"
+    )
+    assert received_df.loc[0, "sender_peer_id"] == "16U*wJXtuH"
+    assert received_df.loc[0, "timestamp"] == pd.Timestamp(1763643976019361536, unit="ns")
+    assert pd.api.types.is_datetime64_ns_dtype(received_df["timestamp"])
+
+
+def test_received_regex_matches_json():
+    tracer = WakuTracer().with_received_pattern_group()
+
+    line = (
+        '{"lvl":"DBG",'
+        '"ts":2026-07-16 15:10:00.145+00:00,'
+        '"msg":"received relay message",'
+        '"topics":"waku relay",'
+        '"tid":7,'
+        '"file":"protocol.nim:221",'
+        '"my_peer_id":"16U*S6zwhq",'
+        '"msg_hash":"0x4d19ab4d7deebcd5c1935d7c27caab678566dbb14f3bf7858d09d466e7c79af7",'
+        '"msg_id":"b9cfa467b62a...67128d31aecc",'
+        '"from_peer_id":"16U*74x9yd",'
+        '"topic":"/waku/2/rs/0/0",'
+        '"contentTopic":"/my-app/1/dst/proto",'
+        '"receivedTime":1784214600145432320,'
+        '"payloadSizeBytes":1000.0}'
+    )
+
+    parsed_logs = mock_queries([[line]], tracer)
+    dfs = tracer.trace(parsed_logs)
+    received_df = dfs["received"][0]
+
+    assert list(received_df.columns[:4]) == [
+        "receiver_peer_id",
+        "msg_hash",
+        "sender_peer_id",
+        "timestamp",
+    ]
+    assert received_df.loc[0, "receiver_peer_id"] == "16U*S6zwhq"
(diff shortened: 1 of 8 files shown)


src/deployments/experiments/waku.py
```diff
@@ -88,59 +160,89 @@ def log_event(self, event):
...
-        namespace = nodes["metadata"]["namespace"]
-        logger.info(f"Starting disconnect+publish loop for nodes in `{name}`")
+        relay_nodes = deployments["relay"]
+        num_relay_nodes = relay_nodes.spec.replicas
+        relay_name = relay_nodes.metadata.name
+        bootstrap_name = deployments["bootstrap"].metadata.name
+        namespace = relay_nodes.metadata.namespace
+        logger.info(f"Starting disconnect+publish loop for nodes in `{relay_name}`")
+
+        if grab_metrics:
```
AlbertoSoutullo: Isn't this checking the function `grab_metrics` instead of `self.config.grab_metrics`?

src/deployments/experiments/waku.py
```diff
@@ -20,37 +24,93 @@
...
+            pod_name,
+            ["/bin/sh", "-c", f"curl -s localhost:{metrics_port}/metrics"],
+        )
+        await command.collect_output_async()
+        if not command.ok:
+            raise RuntimeError(f"Command failed: {command}")
+        logger.debug(f"Retreived /metrics from pod: `{pod_name}` response: `{command.output}`")
+        return command.output
+    except client.exceptions.ApiException as e:
+        logger.error(f"Exception when grabbing /metrics: {e}")
```
AlbertoSoutullo: Is this expected to catch an exception and return None? Because then this is passed to `self.dump` and it will rise a TypeError. Should be handled.

src/deployments/experiments/multi_waku.py
```diff
@@ -0,0 +1,117 @@
...
+                            configs.append(config_params)
+
+        return configs
+
+    def get_params_list(self) -> List[dict]:
+        return [item for item in self.exp_params()]
+
+    def get_name_from_params(self, params: dict) -> str:
+        version = params["image"].tag
+        keys = ["num_nodes", "num_messages"]
```
AlbertoSoutullo: Hmm doesn't `exp_params()` produce "num_relay_nodes", "delay_after_publish", "msg_size_kbytes", "muxer", and other parameters. It doesn't match with "num_nodes".

AlbertoSoutullo (changes requested):
I saw some weird things

src/deployments/experiments/waku.py
PearsonWhite: We print an error log and ignore the error. `_dump` will just write `None`, which is accurate.

src/deployments/experiments/waku.py
PearsonWhite: Fixed in 86cc8a8dc8680e31a557c3953045d7439978a398

src/deployments/experiments/multi_waku.py
PearsonWhite: Fixed in 86cc8a8dc8680e31a557c3953045d7439978a398

src/deployments/experiments/waku.py
AlbertoSoutullo: I don't agree with that. Unless I am missing something, `dump_metrics()` passes this `None` value to `dump()`, which calls `out_file.write(obj)`. A metrics request failure then will raise `TypeError` because that call is not correctly protected by this.
So we either return an empty string/explicit failure and skip the file, or catch the exception in `_dump()`/the pre-publish call. 
Also, the function’s return annotation should also reflect failure, ie: `Optional[str]`.

src/deployments/experiments/multi_waku.py
```diff
@@ -0,0 +1,117 @@
...
+    def get_params_list(self) -> List[dict]:
+        return [item for item in self.exp_params()]
+
+    def get_name_from_params(self, params: dict) -> str:
+        version = params["image"].tag
+        keys = ["num_relay_nodes"]
+        used_keys = filter(lambda item: item[0] in keys, params.items())
+        param_list = [f"{key}_{value}" for key, value in used_keys]
+        param_list.append(f"version_{version}")
+        return "__".join(param_list)
```
AlbertoSoutullo, line 117: Will this not interfiere/conflict when a Multiple experiment is launched?

src/analysis/mesh_analysis/readers/tracers/tests/test_waku_tracer.py
```diff
@@ -0,0 +1,90 @@
...
+    assert received_df.loc[0, "timestamp"] == pd.Timestamp(1763643976019361536, unit="ns")
+    assert pd.api.types.is_datetime64_ns_dtype(received_df["timestamp"])
+
+
+def test_received_regex_matches_json():
+    tracer = WakuTracer().with_received_pattern_group()
+
+    line = (
+        '{"lvl":"DBG",'
+        '"ts":2026-07-16 15:10:00.145+00:00,'
```
AlbertoSoutullo: This has an unquoted timestamp. The test exercises the regex, not JSON parsing, so this does not currently fail for the right reason.

AlbertoSoutullo (changes requested):

src/deployments/experiments/multi_waku.py
PearsonWhite: No, because `src/deployments/experiments/multi_experiment.py` add the random number at the end.

src/analysis/mesh_analysis/readers/tracers/tests/test_waku_tracer.py
PearsonWhite: Fixed in 4db83b423af193ca5aa25daf60a7b5227dd4cc4e

src/deployments/experiments/waku.py
PearsonWhite: Moved the try/catch to the caller in 4db83b423af193ca5aa25daf60a7b5227dd4cc4e

src/analysis/mesh_analysis/readers/tracers/waku_tracer.py
radiken: Small gap in the json support here, and line 63 below, they still use bare `=` so I think under `--log-format=JSON` they won't match.

PearsonWhite: Updated in 290d34cd042b7c9506433e9fb6b9b067f291b65e

src/deployments/experiments/waku.py
```diff
@@ -88,59 +157,92 @@ def log_event(self, event):
...
+    ):
+        folder_prefix = Path(folder_prefix) if folder_prefix else Path("")
+
+        async def _dump(node_index: int):
+            logger.debug(f"Grabbing metrics for {node_name}-{node_index}")
+            try:
+                metrics = await grab_metrics(self.namespace, f"{node_name}-{node_index}")
+                self.dump(metrics, folder_prefix / f"metrics_{node_name}-{node_index}.log")
+            except client.exceptions.ApiException as e:
+                logger.error(f"Exception when grabbing /metrics: {e}")
```
AlbertoSoutullo, line 177: The except is `ApiException` but `grab_metrics()`raises `RuntimeError`.

AlbertoSoutullo (changes requested):

radiken (approved):
LGTM except the unresolved comments

src/deployments/experiments/waku.py
PearsonWhite: Fixed in 1fd2a6cd8f2780a13b9072dfdc29c44642c2b685

AlbertoSoutullo (approved):

PR #346, opened 2026-07-09
Kad-dht regression sweep across versions and muxers

radiken:
Adds `multi_nimlibp2p_kad`, a `Multiple` experiment that runs the nim-libp2p regression as a kad-dht sweep.

### DIFF
--- a/src/deployments/experiments/libp2p/multi_nimlibp2p_kad.py
+++ b/src/deployments/experiments/libp2p/multi_nimlibp2p_kad.py
@@ -1,0 +1,75 @@
+import os
+from typing import Iterable, List, Optional
+
+from src.deployments.core.configs.container import Image
+from src.deployments.experiments.libp2p.nimlibp2p import NimLibp2pExperiment
+from src.deployments.experiments.multi_experiment import Multiple
+from src.deployments.registry import experiment
+
+# nim-libp2p regression matrix: each version x muxer is one full kad-dht run.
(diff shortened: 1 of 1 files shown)


src/deployments/experiments/libp2p/multi_nimlibp2p_kad.py
```diff
@@ -0,0 +1,75 @@
...
+from src.deployments.experiments.multi_experiment import Multiple
+from src.deployments.registry import experiment
+
+# nim-libp2p regression matrix: each version x muxer is one full kad-dht run.
+# The sweep is env-configurable so the same experiment drives a smoke test, a
+# single-version pass (runs are sequential to keep measurements clean), or the
+# full matrix -- see REGRESSION_* below.
+IMAGES = {
+    "2.0.0": Image(repo="radiken/dst-test-node-regression", tag="v2.0.0-kad"),
+    "2.1.0": Image(repo="radiken/dst-test-node-regression", tag="v2.1.0-kad"),
```
AlbertoSoutullo, line 15: Just a question, kad should not be an image right? Wasn't the idea to make this toggable?

radiken: It is toggable with the same image. The `-kad` tag here is just the old tag name for the 2.0.0/2.1.0 regression builds, 2.2.0 already dropped it.

AlbertoSoutullo (approved):

PR #348, opened 2026-07-12
Add tests for multi_experiment

mamoutou-diarra:
- Add new tests for multi_experiments

### DIFF
--- a/src/deployments/experiments/tests/test_multi_experiment.py
+++ b/src/deployments/experiments/tests/test_multi_experiment.py
@@ -1,0 +1,202 @@
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+from pydantic import PrivateAttr
+
+from src.deployments.experiments.multi_experiment import Config, Multiple, dict_to_namespace
+from src.deployments.registry import registry as experiment_registry
+
+# --------------------------------------------------------------------------- #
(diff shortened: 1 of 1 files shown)


AlbertoSoutullo (approved):

radiken (approved):
LGTM!

src/deployments/experiments/tests/test_multi_experiment.py
```diff
@@ -0,0 +1,202 @@
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+from pydantic import PrivateAttr
+
+from src.deployments.experiments.multi_experiment import Config, Multiple, dict_to_namespace
+from src.deployments.registry import registry as experiment_registry
```
PearsonWhite: We're using a global mutable `experiment_registry` here because `src/deployments/experiments/multi_experiment.py` uses it. But if we run tests in parallel, they may interfere with each other. What we could do instead is inject `experiment_registry` into `Multiple` with `mocker.patch`.

mamoutou-diarra: good catch, I've modified `registered_target` to create local registry instance and inject it with patch

PearsonWhite (approved):

PR #349, opened 2026-07-13
DST Dashboard Backend

### LINKED ISSUE #219: Dashboard
[Roadmap Link](https://roadmap.vac.dev/dst/ift/2026q2-dst-tooling#dashboard)

#### Description
Present DST reports automatically in a dashboard for other CCs.

mamoutou-diarra:
## Context
This PR adds the DST Dashboard  **backend** to 10ksim. The dashboard (FastAPI/uvicorn + MongoDB) exposes endpoints to the frontend to create/update/delete experiments, panels, and datasets. Datasources (VictoriaLogs, Prometheus) are provided via `config.yaml`. All write endpoints are JWT-protected. Tokens are generated from an admin page (accessible from the lab homepage, Authentik SSO required) that also provides an experiment CRUD UI for pasting JSON. Data is fetched using DataPuller and scrape_utils, and transform functions convert datasets into ECharts-ready panel specs.

## Main changes
- Added experiment, dataset, and panel processors to fetch data, apply transformations, build ECharts-ready panels, and store them on MongoDB
- Added transform functions for boxplot, timeseries, and utilities like top-N filtering, groupBy, firstN, regex_match
- Implemented a MongoDB storage layer, using GridFS to store large experiment datasets/panels in chunks
- Added CRUD endpoints for experiments, datasets, and panels, exposed via Swagger, with writes gated behind JWT
- Added an admin page for token generation and experiment CRUD (paste JSON)
- Parallelized dataset and panel processing across a thread pool
- Added unit tests for panel processors
- Added script and example for adding new ewperiment json data (They can be added also by curl)

### Deployment
- Added k8s manifests for the backend, frontend, and MongoDB (dev/prod)

## App URLs
- Backend API: [https://api.dashboard.lab.vac.dev](https://api.dashboard.lab.vac.dev)
- Admin Page (token generation, Add/Update/Delete Experiments): [https://api.dashboard.lab.vac.dev/admin/token](https://api.dashboard.lab.vac.dev/admin/token) (require Authentication via keycloak)


dst_dashboard/main.py
```diff
@@ -0,0 +1,86 @@
...
+
+    except Exception as e:
+        logger.error(f"Failed to initialize DST Dashboard: {e}", exc_info=True)
+        sys.exit(1)
+
+
+# Enable CORS for frontend
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
```
AlbertoSoutullo: This combined with JWT tokens in browser sessions allows any origin to make credentialed requests. 
Should it be restricted to the known frontend origin? (ie dashboard.lab.vac.dev).

dst_dashboard/k8s/mongodb-dev.yaml
```diff
@@ -0,0 +1,79 @@
...
+          imagePullPolicy: IfNotPresent
+          ports:
+            - containerPort: 27017
+              name: mongodb-dev
+              protocol: TCP
+          env:
+            - name: MONGO_INITDB_ROOT_USERNAME
+              value: "root"
+            - name: MONGO_INITDB_ROOT_PASSWORD
+              value: "" #password
```
AlbertoSoutullo: I think this should be used as secrets, right?

dst_dashboard/api/datasets.py
```diff
@@ -0,0 +1,151 @@
...
+    # Get experiment from database (source of truth)
+    experiment_data = db.get_experiment(experiment_id)
+    if experiment_data is None:
+        raise HTTPException(status_code=404, detail="Experiment not found")
+
+    experiment = ExperimentConfig(**experiment_data)
+
+    # Initialize database
+    db = DSTDatabase()
+
```
AlbertoSoutullo: `db = DSTDatabase()` is called twice here

dst_dashboard/api/admin.py
```diff
@@ -0,0 +1,157 @@
...
+                "status": "success",
+                "message": f"Experiment '{experiment_id}' reprocessed successfully",
+                "experiment_id": experiment_id,
+                "datasets_count": len(experiment.datasets),
+                "panels_count": len(experiment.panels),
+            }
+        else:
+            raise HTTPException(
+                status_code=500, detail=f"Failed to reprocess experiment '{experiment_id}'"
+            )
```
AlbertoSoutullo: I think there is a bug in here.
`processor.process_experiment(experiment)` is supposed to return a str that represents the experiment ID.
So either it returns the experiment ID and this is used, or it returns a bool. But right now this is confusing.

dst_dashboard/processors/panel_processor.py
```diff
@@ -0,0 +1,500 @@
...
+                    "name": panel_config.title,
+                    "type": "line",
+                    "data": y_values,
+                    "smooth": True,
+                    "sampling": "lttb",
+                    "symbol": "none",
+                    "lineStyle": {"width": 2, "color": "#00d4ff"},
+                    "areaStyle": {"color": "#00d4ff", "opacity": 0.15},
+                }
+            ]
```
AlbertoSoutullo, line 349: When no `groupBy` is set, series data is a plain list of y-values ([1.0, 2.0]). But the grouped path produces [[timestamp, value], ...] pairs. This inconsistency is even noted in the test (test_single_series_without_groupby_has_no_x_pairing), but the x-axis type is "time", so ECharts will misrender the single-series case. The single-series path should also produce [x_val, y_val] pairs.

dst_dashboard/api/datasets.py
```diff
@@ -0,0 +1,151 @@
...
+        data = processor.fetch_dataset(experiment_id, dataset_config)
+
+        # Update cache
+        if data:
+            db.store_dataset(experiment_id, dataset_name, data)
+
+        return {"data": data, "source": "datasource"}
+
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=f"Failed to fetch dataset: {str(e)}")
```
AlbertoSoutullo: I don't know if I am comfortable with a cached data by default. Maybe the re-refresh option should be True just in case?

dst_dashboard/processors/dataset_processor.py
```diff
@@ -0,0 +1,374 @@
...
+            return tracer
+        else:
+            # Fallback to generic MessageTracer
+
+            tracer = MessageTracer()
+            if pattern:
+                tracer.with_wildcard_pattern()
+            else:
+                tracer.with_wildcard_pattern()
+            return tracer
```
AlbertoSoutullo, line 190: Uhmm this if else is the same.

AlbertoSoutullo (changes requested):
I think this could have been splitted even more into smaller PRs. Consider that for next time please.
I would also appreciate an addition about how the flow of this works to better understand the modules.

dst_dashboard/main.py
mamoutou-diarra: > This combined with JWT tokens in browser sessions allows any origin to make credentialed requests. Should it be restricted to the known frontend origin? (ie dashboard.lab.vac.dev).

Yes, we could restrict it to our frontend domain for strict security.

dst_dashboard/k8s/mongodb-dev.yaml
mamoutou-diarra: > I think this should be used as secrets, right?

Yes, this is used as secret during the deployment, I'll change it here to make it consistent

dst_dashboard/api/datasets.py
mamoutou-diarra: Thanks for flagging this. Actually, after your comment, I realized that refresh will make panels stale so I'll remove the refresh option here. Panels are cached separately and are only regenerated when the admin reprocess endpoint is called, so a dataset-only refresh would leave related panels  stale. To avoid that I made reprocessing the single way to get fresh data (in other words after calling the reprocessing endpoint, you can call GET on `/experiments/..../datasets/<name>/data` to get fresh dataset.

dst_dashboard/api/admin.py
mamoutou-diarra: fixed now

dst_dashboard/processors/panel_processor.py
mamoutou-diarra: good catch, this is indeed a very bad issue. I fixed it and updated the tests

dst_dashboard/api/datasets.py
mamoutou-diarra: fixed now

dst_dashboard/processors/dataset_processor.py
mamoutou-diarra: fixed now, thanks

AlbertoSoutullo (approved):

PR #351, opened 2026-07-16
Shadow: scrape gossipsub control/efficiency metrics

radiken:
Adds a gossipsub-detail metric set (IHAVE/IWANT/GRAFT/PRUNE sent+received, duplicates, received, IDONTWANT-saved) to the Shadow scrape, reduced per node into across-node medians + a duplicate ratio. The data is already in every node's `/metrics` dump; this just queries and summarizes it.


AlbertoSoutullo (approved):
Files like `src/analysis/metrics/libp2p/metrics.py` always bug me a bit but I don't see another cleaner way to do it tbh... So LGTM

PR #352, opened 2026-07-16
k8s: scrape gossipsub detail from cluster regression runs

radiken:
Wires the gossipsub-detail scrape (from #351) into the cluster scrape path, so cluster regressions capture the same IHAVE/IWANT/GRAFT/PRUNE + duplicate-ratio detail as Shadow. The cluster VM already keeps these series, so no vaclab change is needed. Stacked on #351.

### DIFF
--- a/scrape.py
+++ b/scrape.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Iterable, Union
 
 from src.analysis.metrics.config import ScrapeConfig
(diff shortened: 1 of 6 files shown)


AlbertoSoutullo (commented):
Looks good to me but I think some of those functions could have tests to support the PR a bit.

AlbertoSoutullo (approved):

PR #316, opened 2026-07-03
shadow-gossipsub: select muxer and discovery (static | kad-dht)

radiken:
The shadow experiment can only run yamux in static mode currently, this PR adds muxer and discovery selection.

- `render_shadow_yaml` gains `discovery` and `start_sleep`.
- `shadow-gossipsub` ExpConfig exposes `muxer`, `discovery`, `start_sleep` and passes them through.
- `multi-shadow-gossipsub` is now env-configurable (`SHADOW_MUXERS`/`SHADOW_SIZES`/`SHADOW_DISCOVERY`).
- kad-dht bootstrap connection cap is lifted above the network size (otherwise only 250 nodes bootstrap), and matrix run names are shortened so the log-reader pod stays under the 63-char name limit.
- new determinism/diagnostics/quic knobs: `seed`, `strace_logging_mode`, `model_unblocked_syscall_latency`, `lsquic_tick_floor_us` (needs the tick-floor node image), `start_jitter_ms`.

Validated at 1000 nodes (kad-dht) against the cluster regression: mplex and yamux both hit 100% delivery. quic is now solved as well: the freeze was a zero-delay tick livelock in the lsquic wrapper (fixed by the tick-floor image + `lsquic_tick_floor_us`) plus simultaneous-dial collisions from lockstep starts (fixed by `start_jitter_ms`) - quic@1000 completes in ~1h25m at 100% delivery (600000/600000).

PR #353, opened 2026-07-16
Shadow WAN network model

radiken:
Adds `latency_ms` / `bandwidth_mbit` knobs that replace the ~0-latency `1_gbit_switch` with a one-node GML link, so Shadow runs can model a realistic WAN. Default keeps the switch unchanged. Mirrors the existing k8s network-shaping (netem delay + tbf bandwidth); packet loss omitted on both for now. Stacked on #316.

### DIFF
--- a/src/deployments/shadow/builders.py
+++ b/src/deployments/shadow/builders.py
@@ -147,6 +147,10 @@ def _publisher_host(publisher_start_s: int, requester_app_path: str) -> dict:
     }
 
 
+_SWITCH_BANDWIDTH_MBIT = 1000
+_SWITCH_LATENCY_MS = 0
+
+
 def _wan_gml(latency_ms: int, bandwidth_mbit: int) -> LiteralScalarString:
     """A one-node GML graph with a self-loop, so every host (all on network_node_id 0)
     sees `latency_ms` one-way delay and `bandwidth_mbit` up/down. Shadow's GML parser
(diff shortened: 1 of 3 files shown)


src/deployments/shadow/builders.py
```diff
@@ -194,16 +219,24 @@ def render_shadow_yaml(
...
+
+    # Network: the built-in 1_gbit_switch is ~0-latency, which is unrealistic and can
+    # skew latency-sensitive behaviour. Set latency_ms (or a sub-1Gbit bandwidth) to model
+    # a WAN via a one-node GML with a self-loop (all hosts share network_node_id 0), the
+    # same shape the standalone shadow harness used.
+    if latency_ms > 0 or bandwidth_mbit < 1000:
+        graph = {"type": "gml", "inline": _wan_gml(latency_ms, bandwidth_mbit)}
+    else:
+        graph = {"type": "1_gbit_switch"}
+
```
AlbertoSoutullo, line 241: Is there a particular reason why the if statement is `latency_ms > 0 or bandwidth_mbit < 1000`?
I think it would be nice to have a check wether the value is negative, and If it has to be lower than 1000 then don't accept those values. Because if you pass a higher value now, it is ignored.

AlbertoSoutullo (changes requested):

src/deployments/shadow/builders.py
radiken: Good catch! Thanks! Both are None by default now. Also added tests for bandwidth above 1000, negatives, and zero.

AlbertoSoutullo (approved):

PR #354, opened 2026-07-16
Fix command add_args

PearsonWhite:
Add_args was comparing the flag to tuples when searching for existing args, thus failing to properly replace when on_duplicate="replace".


radiken (approved):
LGTM!

AlbertoSoutullo (approved):
LGTM

PR #355, opened 2026-07-17
Frontend Enhancements - 1

mamoutou-diarra:
## Context
This PR polishes several parts of the UI and fixes some identified issues

## Major changes
- Fixed the experiment detail page's back link to return to the Experiments list instead of Home
- Fixed page scroll position/layout jumping when navigating between pages
- Fixed the "vaclab" navbar dropdown so Network/Resources can actually be clicked on hover
- Removed the Twitter/X link, replaced with our Discord channel (navbar + footer)
- Renamed "VacLab" to "vaclab", reworded Experiments page text, renamed footer "Product" section to "Explore"
- Updated vaclab GitHub links to point to the actual repo instead of the org root
- Added a "Latest experiments" section to the homepage (6 most recent), each with an auto-rotating chart-panel preview
- Enabled scroll-triggered fade-in animations to show homepage content progressively as we scroll down
- Redesigned the Experiments page filter bar (replaced category text and added a search bar) and fixed inconsistent card heights
- Refined light/dark theme consistency (corner radius, typography)
- Replaced plain loading text with a spinner
## URLs
- Already accessible in DEV: [https://dev.dashboard.lab.vac.dev](https://dev.dashboard.lab.vac.dev)

### DIFF
--- a/dst_dashboard/frontend/src/components/ExperimentThumbnail.js
+++ b/dst_dashboard/frontend/src/components/ExperimentThumbnail.js
@@ -1,0 +1,108 @@
+import React, { Suspense, lazy, useEffect, useState } from 'react';
+import axios from 'axios';
+import { API_BASE_URL } from '../config';
+import { useInView } from '../hooks/useInView';
+import { buildThumbnailOption } from '../utils/chartOptions';
+import { useTheme } from '../context/ThemeContext';
+
+const ReactECharts = lazy(() => import('echarts-for-react'));
+
+const THUMB_HEIGHT = 152;
+const MAX_PANELS = 4;
+const ROTATE_INTERVAL_MS = 15000;
+
+function ThumbnailSkeleton() {
+    return (
+        <div className="animate-pulse w-full" style={{ height: THUMB_HEIGHT }}>
+            <div className="h-full w-full bg-base-300/60" />
+        </div>
+    );
+}
+
+// A quick "glance" preview of an experiment's results, rendered only once the
+// card scrolls near the viewport. Fetches up to MAX_PANELS panels once (not
+// re-fetched on re-scroll - useInView's default `once` gating), then cycles
+// through whichever came back, crossfading every few seconds so a card with
+// several panels doesn't just sit on the first one forever.
+function ExperimentThumbnail({ experimentId }) {
+    const { isDark } = useTheme();
+    const [ref, inView] = useInView();
+    const [options, setOptions] = useState([]);
+    const [index, setIndex] = useState(0);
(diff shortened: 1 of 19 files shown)


AlbertoSoutullo (changes requested):
Reviewing with AI. There are some points that I think they are not really super important right now but there is one that maybe it can be concerning. In the homepage, we fetch up to the latest 6 experiments. I think the preview in the homepage should be something we have in cache actually.

PearsonWhite (approved):
Good incremental change.

dst_dashboard/frontend/src/components/ExperimentThumbnail.js
```diff
@@ -0,0 +1,108 @@
...
+                }
+            })
+            .catch(() => {
+                if (!cancelled) setFailed(true);
+            });
+
+        return () => {
+            cancelled = true;
+        };
+    }, [inView, experimentId, isDark]);
```
radiken: `isDark` here re-runs the whole fetch on every theme toggle, but it's only consumed by `buildThumbnailOption`. I guess we can keep the raw options in state and derive the themed ones in a `useMemo` on `isDark`, leaving this effect on `[inView, experimentId]`.

radiken (commented):
LGTM with just one small optional improvement

dst_dashboard/frontend/src/components/ExperimentThumbnail.js
mamoutou-diarra: > `isDark` here re-runs the whole fetch on every theme toggle, but it's only consumed by `buildThumbnailOption`. I guess we can keep the raw options in state and derive the themed ones in a `useMemo` on `isDark`, leaving this effect on `[inView, experimentId]`.

Thanks for this nice suggestion, I fixed it: Toggling the theme now just recomputes in memory instead of re-fetching the experiment + panel data.

AlbertoSoutullo (commented):
Can it be that right now we are caching partial results?
If the user navigates away while only one of several panel requests has completed, the cache contains a partial result. When the homepage is mounted again, that partial array is treated as a complete cache entry, so the remaining panels are never requested.

AlbertoSoutullo (changes requested):
Last comment (I swear) we are doing 6 cards × 4 panels per card = 24 panel requests
I would reduce that to 6 total, because if not we are fetching an insane amount of data

AlbertoSoutullo (changes requested):
The PR description says homepage thumbnails should have an “auto-rotating chart-panel preview.”. The current component fetches only one successful panel and renders it permanently. Which one are we sticking to?

PR #356, opened 2026-07-17
k8s: optional bandwidth cap on the network-delay init container

radiken:
Adds a `network_bandwidth_mbit` knob to the nimlibp2p experiment. Off by default (0 = uncapped), so existing runs are unchanged.

### DIFF
--- a/src/deployments/core/configs/helpers/utils.py
+++ b/src/deployments/core/configs/helpers/utils.py
@@ -1,6 +1,6 @@
 # Python Imports
 from copy import deepcopy
-from typing import Dict, List, Literal, Optional, Tuple, Type, TypeVar, Union, get_args
+from typing import Dict, List, Literal, Optional, Tuple, Type, TypeVar, get_args
 
 from kubernetes.client import V1Capabilities, V1Container, V1SecurityContext
 from pydantic import NonNegativeInt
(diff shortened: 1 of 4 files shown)


src/deployments/core/configs/helpers/utils.py
```diff
@@ -164,15 +164,22 @@ def convert_to_container_config(
...
+        netem += f" rate {rate_mbit}mbit"
     return V1Container(
         name="slowyourroll",
         image="soutullostatus/tc-container:1",
         image_pull_policy="IfNotPresent",
         security_context=V1SecurityContext(capabilities=V1Capabilities(add=["NET_ADMIN"])),
-        command=[
-            f"tc qdisc add dev eth0 root netem delay {str(delay)}ms {str(jitter)}ms distribution normal",
-        ],
+        command=[f"tc qdisc add dev eth0 root netem {netem}"],
```
AlbertoSoutullo, line 179: `delay` and `jitter` accept `str`, but the implementation always appends ms, so "100ms" becomes "100msms". We have callers and tests using strings with units, even though the type suggests numbers.
I think it would be good to normalize things and use numbers.

AlbertoSoutullo (changes requested):
Minor comment

src/deployments/core/configs/helpers/utils.py
radiken: Thanks! Fixed now

AlbertoSoutullo (approved):

PR #358, opened 2026-07-21
Regression analysis: plot delivery latency as a CDF

radiken:
Reads the delays the reliability analysis already dumps to `analysis_data/summary/received.csv` and draws a CDF per run, plus a percentile table for the report.


AlbertoSoutullo (approved):
I am approving even tho the function is not called anywhere because it will be done in an upcoming PR?

PR #359, opened 2026-07-21
nimlibp2p: outlast a metrics scrape before teardown

radiken:
Cleanup ran 20s after the last publish, so pods whose final scrape landed before then had their counters frozen mid-run. Dwell 90s by default, configurable.


AlbertoSoutullo (approved):
Thanks!

PR #362, opened 2026-07-23
Generate latency plots from the regression pipeline

radiken:
Wires the existing LatencyPlotter into the cluster regression flow and adds a CLI so Shadow runs use the same component.


AlbertoSoutullo (commented):
Hmmm but this only happens when we run `log_multi_analysis.py` right? Experiments using the post-run hoot will not be able to use this. I think we should move this to a post-run analysis implementation.

AlbertoSoutullo (approved):

PearsonWhite (approved):

PR #363, opened 2026-07-23
Allow non-primitive objects in ExpConfig CLI args

PearsonWhite:
Change auto generation of experiment config CLI params to allow passing in complex types.

For types that inherit from pydantic.BaseModel, allow JSON str input. 
```--image='{"repo":"pearsonwhite/nwaku","tag":"quic-b778d16"}'```

For types that have a from_str static method, use that if JSON parsing fails.
```--image="pearsonwhite/nwaku:quic-b778d16"```

Falls back on direct constructor with input str as the argument.

### DIFF
--- a/src/deployments/utils/parser.py
+++ b/src/deployments/utils/parser.py
@@ -1,6 +1,7 @@
 # Python Imports
 import argparse
 import json
+import types
 from typing import Annotated, Any, Literal, Union, get_args, get_origin
 
 from pydantic import BaseModel, ValidationError
@@ -27,20 +28,23 @@ def _annotation_display(annotation) -> str:
 
 
 def _unwrap_optional(annotation):
+    """Recursively unwrap Annotated and Optional to get the actual type."""
     origin = get_origin(annotation)
 
-    # Unwrap Annotated[T, ...] -> T
+    # Unwrap Annotated[T, ...] -> T, then recurse
     if origin is Annotated:
         args = get_args(annotation)
         if args:
-            annotation = args[0]
-            origin = get_origin(annotation)
+            return _unwrap_optional(args[0])
 
     # Unwrap Optional[T] -> T
-    if origin is Union:
+    # Optional[T] => typing.Union
+    # T | None => types.UnionType
+    if origin is Union or origin is types.UnionType:
         args = [arg for arg in get_args(annotation) if arg is not type(None)]
         if len(args) == 1:
-            return args[0]
+            return _unwrap_optional(args[0])
+
     return annotation
 
 
--- a/src/deployments/utils/parser.py
+++ b/src/deployments/utils/parser.py
@@ -1,6 +1,7 @@
 # Python Imports
 import argparse
-from typing import Any, Literal, Union, get_args, get_origin
+import json
+from typing import Annotated, Any, Literal, Union, get_args, get_origin
 
 from pydantic import BaseModel, ValidationError
 from pydantic.fields import FieldInfo
@@ -27,6 +28,15 @@ def _annotation_display(annotation) -> str:
 
 def _unwrap_optional(annotation):
     origin = get_origin(annotation)
+
+    # Unwrap Annotated[T, ...] -> T
+    if origin is Annotated:
+        args = get_args(annotation)
+        if args:
+            annotation = args[0]
+            origin = get_origin(annotation)
+
+    # Unwrap Optional[T] -> T
     if origin is Union:
         args = [arg for arg in get_args(annotation) if arg is not type(None)]
         if len(args) == 1:
@@ -36,10 +46,11 @@ def _unwrap_optional(annotation):
 
 def get_from_str(annotation: Any, field_name: str):
     def from_str(input_str: str) -> Any:
-        # Convert using model_validate_json.
         errors = []
+
+        # Convert using model_validate_json.
         try:
-            if issubclass(annotation, BaseModel):
+            if isinstance(annotation, type) and issubclass(annotation, BaseModel):
                 return annotation.model_validate_json(input_str)
         except (ValidationError, ValueError) as e:
             err = argparse.ArgumentTypeError(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/deployments/utils/parser.py [hunk 1]
```diff
@@ -63,6 +106,10 @@ def _field_to_arg(field_name: str, field: FieldInfo) -> tuple[str, dict[str, Any
             kwargs["type"] = type(choices[0])
             kwargs["choices"] = choices
 
+    if "type" not in kwargs.keys() and annotation is not bool:
+        from_str = get_from_str(annotation, field_name)
+        kwargs["type"] = from_str
```
radiken: This catches containers too, `annotation(input_str)` succeeds on them because `list("peerA")` is valid in Python. Could do something like this instead:


```suggestion
    if "type" not in kwargs and annotation is not bool:
        if origin is not None or annotation in (list, set, dict, tuple):
            kwargs["type"] = str
        else:
            kwargs["type"] = get_from_str(annotation, field_name)
```

radiken (changes requested):

src/deployments/utils/parser.py
```diff
@@ -33,6 +34,49 @@ def _unwrap_optional(annotation):
...
+        # Convert using model_validate_json.
+        errors = []
+        try:
+            if issubclass(annotation, BaseModel):
+                return annotation.model_validate_json(input_str)
+        except (ValidationError, ValueError) as e:
+            err = argparse.ArgumentTypeError(
+                f"Failed convert string with model_validate_json. Field: `{field_name}` Error: `{e}`"
+            )
+            errors.append(err)
```
AlbertoSoutullo: Some typing annotations and special typing objects are not valid first arguments to `issubclass`, and can raise `TypeError` rather than returning `False`.

For example, fields annotated with Any or other non-class typing may reach this code because they have no origin and are not bool, int, float, or str. In that case, parsing the CLI argument can fail with an uncaught `TypeError` instead of an `argparse.ArgumentTypeError`.

src/deployments/utils/parser.py
```diff
@@ -33,6 +34,49 @@ def _unwrap_optional(annotation):
...
+        except AttributeError as e:
+            err = argparse.ArgumentTypeError(
+                f"No from_str method for object. Field: `{field_name}` Error: `{e}`"
+            )
+            errors.append(err)
+        except (ValidationError, ValueError, TypeError) as e:
+            err = argparse.ArgumentTypeError(
+                f"Failed to convert string with from_str. Field: `{field_name}` Error: `{e}`"
+            )
+            errors.append(err)
```
AlbertoSoutullo: This block treats any AttributeError raised by `annotation.from_str(input_str)` as meaning that the method does not exist.
I think we should check for the method before calling it, so `AttributeErrors` makes sense instead of reporting a missing method.

AlbertoSoutullo (changes requested):

src/deployments/utils/parser.py
PearsonWhite: >so AttributeErrors makes sense instead of reporting a missing method.

~~What do you mean? We should check it, then try it, then report the AttributeError?~~

edit: Oh I see. You're saying from_str could have an error inside of it that bubbles up an AttributeError. Ok.

PearsonWhite: Addressed in ecf92cbd621dc2c404c4252493c34e71da5db4ee

src/deployments/utils/parser.py
PearsonWhite: Addressed in ecf92cbd621dc2c404c4252493c34e71da5db4ee

src/deployments/utils/parser.py
AlbertoSoutullo: Yes, exactly. I would separate method lookup from method execution

AlbertoSoutullo: Please:
- Separate from_str lookup from invocation.
- Add a test for a failing from_str.
- Decide explicitly whether internal AttributeError should become an argparse error or propagate as a bug.

The function currently does not clearly distinguish “the conversion method is absent” from “the conversion method exists but failed.”

AlbertoSoutullo (changes requested):

src/deployments/utils/parser.py
PearsonWhite: 884b9af87d13818be8c203406cd082392e7e43fb
1. I removed the errors from attribute not existing or constructor not callable, so methods that aren't available will simply be skipped instead of adding to the err output
2. Added test_from_str_falls_back_on_failure
3. All errors from `from_str` should still allow fallback on using the constructor. And all errors from the constructor should still be collected the same way. So I changed it to catch Exception instead of the specific Exception types.

radiken (approved):
LGTM

src/deployments/utils/parser.py
```diff
@@ -26,13 +28,66 @@ def _annotation_display(annotation) -> str:
...
+        if args:
+            annotation = args[0]
+            origin = get_origin(annotation)
+
+    # Unwrap Optional[T] -> T
     if origin is Union:
         args = [arg for arg in get_args(annotation) if arg is not type(None)]
         if len(args) == 1:
             return args[0]
     return annotation
```
AlbertoSoutullo, line 48: `_unwrap_optional()` unwraps `Annotated`, but it does not recursively unwrap the resulting `Optional/Union`.

src/deployments/utils/parser.py
```diff
@@ -26,13 +28,66 @@ def _annotation_display(annotation) -> str:
...
+
+    # Unwrap Annotated[T, ...] -> T
+    if origin is Annotated:
+        args = get_args(annotation)
+        if args:
+            annotation = args[0]
+            origin = get_origin(annotation)
+
+    # Unwrap Optional[T] -> T
     if origin is Union:
```
AlbertoSoutullo: Annotations can write using syntax such as `ImageConfig | None`

AlbertoSoutullo (changes requested):

src/deployments/utils/parser.py
PearsonWhite: Fixed in 518ec03b90953dcc503d5862a892e4cc748df985

src/deployments/utils/parser.py
PearsonWhite: Fixed in 518ec03b90953dcc503d5862a892e4cc748df985

AlbertoSoutullo (approved):

PR #364, opened 2026-07-24
Alberto/improve readmes

AlbertoSoutullo:


src/analysis/README.md
```diff
@@ -1,138 +1,157 @@
-# Distributed deployer framework (Template name)
+# Analysis
 
-**Python Version**: 3.11.5
+`src/analysis` contains the code that turns completed experiment metadata,
+logs, and metrics into checks, intermediate data, plots, and summaries.
 
-## Yaml config:
+The preferred path is automatic post-run analysis:
```
PearsonWhite, line 6: Why is this preferred?

src/analysis/README.md
```diff
@@ -1,138 +1,157 @@
...
+`stack_for_window()` returns a copy of `metadata["stack"]` with
+`start_time`/`end_time` replaced by the selected
+`metadata["results"][window_name]` values. It does not mutate
+`experiment.metadata`.
+
+## DataPuller
+
+`DataPuller` receives the stack and chooses how data is retrieved.
+
+For VictoriaLogs, the stack normally needs:
```
PearsonWhite, line 110: We should decide if we want the adaptors for this that translate metadata or experiment to the stack variables for the anlayzer. Currently in master we have `StackGen` in `log_multi_analysis.py`, but we could also do what is proposed in https://github.com/vacp2p/10ksim/pull/296

If we use one of those, we could reference the `required_fields` in the code here.

src/deployments/README.md
```diff
@@ -1,70 +1,145 @@
...
+    cleanup Kubernetes resources
 
-# Example usage
+write run_finished event
+_dump_metadata()
+run_post_analysis(self)
 ```
-python ./deployment.py -vv --config ~/<kube_config>.yaml regression-nodes --type waku
+
+The experiment's `_run()` method should only do experiment work:
```
PearsonWhite, line 41: I still think we should break this up into: running the experiment (deployments), analysis, and scrape.

PearsonWhite (commented):
Just a few comments for now.

src/analysis/README.md
AlbertoSoutullo: Well, maybe preferred is a bit missleading. Automatic, easier, default maybe are better options

src/analysis/README.md
AlbertoSoutullo: Hmmm I am confused about why this comment is here. Aren't you referring to Bridges and event windows? Or am I mixing things.

src/deployments/README.md
AlbertoSoutullo: Hmmm, what do you mean by this. This readme?

PearsonWhite: I meant in terms of the design of the actual experiment class. Just a comment for later design in the future.

PR #366, opened 2026-07-25
Degraded-network, churn and partition regression scenarios

radiken:
Three adverse-condition variants of the nimlibp2p regression run: shaped links (netem gains a loss knob), killing a share of nodes mid-run, and splitting the network in two then healing it. 

All 3 scenarios are tested at scale.

### DIFF
--- a/src/deployments/core/k8s_cleanup.py
+++ b/src/deployments/core/k8s_cleanup.py
@@ -33,6 +33,7 @@ def get_cleanup_resources(yamls: List[yaml.YAMLObject], types: Optional[List[str
         "Role": [],
         "RoleBinding": [],
         "ServiceAccount": [],
+        "NetworkPolicy": [],
     }
     types = (
         types
@@ -50,6 +51,7 @@ def get_cleanup_resources(yamls: List[yaml.YAMLObject], types: Optional[List[str
             "ConfigMap",
             "ServiceAccount",
             "PersistentVolumeClaim",
+            "NetworkPolicy",
         ]
     )
     for yaml in yamls:
@@ -72,6 +74,12 @@ def delete_pod(name, namespace, *, grace_period=0):
     )
 
 
+def delete_network_policy(name, namespace):
+    client.NetworkingV1Api().delete_namespaced_network_policy(
+        name=name, namespace=namespace, body=client.V1DeleteOptions()
+    )
+
+
 def cleanup_resources(
     resources: dict,
     namespace: str,
@@ -83,6 +91,7 @@ def cleanup_resources(
     """
     logger.info(f"Cleanup resources: `{resources}`")
     deletion_order = [
+        "NetworkPolicy",
         "Deployment",
         "StatefulSet",
         "DaemonSet",
@@ -142,6 +151,9 @@ def cleanup_resources(
         "RoleBinding": lambda name: client.RbacAuthorizationV1Api(
             api_client
         ).delete_namespaced_role_binding(name, namespace, body=client.V1DeleteOptions()),
+        "NetworkPolicy": lambda name: client.NetworkingV1Api(
+            api_client
+        ).delete_namespaced_network_policy(name, namespace, body=client.V1DeleteOptions()),
     }
 
     for kind in deletion_order:
--- a/src/deployments/core/k8s_cleanup.py
+++ b/src/deployments/core/k8s_cleanup.py
@@ -74,8 +74,10 @@ def delete_pod(name, namespace, *, grace_period=0):
     )
 
 
-def delete_network_policy(name, namespace):
-    client.NetworkingV1Api().delete_namespaced_network_policy(
+def delete_network_policy(
+    name: str, namespace: str, api_client: Optional[ApiClient] = None
+) -> None:
+    client.NetworkingV1Api(api_client or client.ApiClient()).delete_namespaced_network_policy(
         name=name, namespace=namespace, body=client.V1DeleteOptions()
     )
 
--- a/src/deployments/core/k8s_rollout.py
+++ b/src/deployments/core/k8s_rollout.py
@@ -2,6 +2,7 @@
 import asyncio
 import logging
 import time
+from concurrent.futures import ThreadPoolExecutor
 from typing import Callable, Iterable, Optional, Tuple
 
 from kubernetes import client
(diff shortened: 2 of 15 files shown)


src/deployments/experiments/libp2p/nimlibp2p.py
```diff
@@ -46,6 +46,7 @@ class ExpConfig(BaseModel):
     network_delay: NonNegativeInt = 0
     network_jitter: NonNegativeInt = 0
     network_bandwidth_mbit: NonNegativeInt = 0  # 0 = uncapped; folded into the netem qdisc
+    network_loss_pct: NonNegativeFloat = 0  # 0 = lossless; folded into the netem qdisc
```
AlbertoSoutullo: We need to add a validator that contraints this value between 0 and 100. It is true it cannot be negative, but also cannot be higher than 100.

src/deployments/experiments/libp2p/regression_scenarios.py
```diff
@@ -0,0 +1,227 @@
...
+
+@experiment(name="nimlibp2p-degraded")
+class DegradedNetwork(NimLibp2pExperiment):
+    """Regression run over a high-latency, jittery, lossy link."""
+
+    config: DegradedConfig
+
+
+class ChurnConfig(ExpConfig):
+    churn_fraction: NonNegativeFloat = 0.2
```
AlbertoSoutullo: Same as before, bound this between 0 and 1.

src/deployments/experiments/libp2p/regression_scenarios.py
```diff
@@ -0,0 +1,227 @@
...
+        await asyncio.sleep(self.config.churn_downtime)
+
+        self.log_event({"event": "churn_rejoin", "nodes": targets})
+        scale_statefulset(name, self.namespace, replicas, self.api_client)
+        await wait_for_rollout(nodes, self.api_client, timeout=self.config.churn_rejoin_timeout)
+        self.log_event({"event": "churn_rejoined", "nodes": targets})
+
+
+class PartitionConfig(ExpConfig):
+    partition_fraction: NonNegativeFloat = 0.5
```
AlbertoSoutullo: Same as others, bound between 0 and 1.

src/deployments/experiments/libp2p/regression_scenarios.py
```diff
@@ -0,0 +1,227 @@
...
+                raise TimeoutError(f"Only {found}/{wanted} pods exist after {timeout}s")
+            await asyncio.sleep(5)
+
+    async def _after_nodes(self, nodes: V1StatefulSet) -> None:
+        side_a, side_b = partition_sides(nodes, self.config.partition_fraction)
+        if not side_a or not side_b:
+            logger.warning(
+                f"partition_fraction {self.config.partition_fraction} leaves one side empty"
+            )
+            return
```
AlbertoSoutullo: That should probably raise a configuration error instead. Otherwise a run named nimlibp2p-partition can complete without testing a partition at all.

src/deployments/core/k8s_rollout.py
```diff
@@ -134,6 +135,32 @@ async def wait_for_rollout(
...
+    with ThreadPoolExecutor(max_workers=32) as pool:
+        return sum(pool.map(patch, names))
+
+
+def scale_statefulset(name: str, namespace: str, replicas: int, api_client=None) -> None:
+    """Set a StatefulSet's replica count. Scaling down removes the highest ordinals."""
+    api = client.AppsV1Api(api_client or client.ApiClient())
+    api.patch_namespaced_stateful_set_scale(
+        name=name, namespace=namespace, body={"spec": {"replicas": replicas}}
+    )
```
AlbertoSoutullo: Does this function wait until the scale down is completed? Because I think we should wait before doing other things

src/deployments/core/k8s_cleanup.py
```diff
@@ -72,6 +74,12 @@ def delete_pod(name, namespace, *, grace_period=0):
     )
 
 
+def delete_network_policy(name, namespace):
+    client.NetworkingV1Api().delete_namespaced_network_policy(
+        name=name, namespace=namespace, body=client.V1DeleteOptions()
+    )
```
AlbertoSoutullo: Most of the new orchestration code passes `self.api_client`, we should do the same here.
Also, typing.

src/deployments/experiments/libp2p/partition.py
AlbertoSoutullo, line 1: Hmm I think we can split this to one file per scenario to make it less bloated

AlbertoSoutullo (changes requested):
Requested some changes but overall it looks good.

src/deployments/experiments/libp2p/regression_scenarios.py
```diff
@@ -0,0 +1,227 @@
...
+                ]
+            )
+        ),
+    ]
+
+
+def build_partition_policy(name: str, namespace: str, side: str, far_side: str) -> V1NetworkPolicy:
+    """Cut `side` off from `far_side`, both directions.
+
+    Ingress alone only stops TCP; quic is UDP and goes straight through. Egress reuses the
```
mamoutou-diarra: AFAIK NetworkPolicy ingress/egress rules with no ports field block all protocols (TCP, UDP, SCTP), not just TCP, so UDP/QUIC is also blocked

mamoutou-diarra (approved):
LGTM, just clarified a misconception about ingress/egress rules with Cilium

mamoutou-diarra (changes requested):
Can you please update the comment about ingress rule?

src/deployments/experiments/libp2p/regression_scenarios.py
radiken: Thanks! Removed that comment now

src/deployments/core/k8s_rollout.py
radiken: It didn't, good catch! Fixed now

src/deployments/experiments/libp2p/nimlibp2p.py
AlbertoSoutullo: This was still not tacled, or I don't see where.

radiken: missed this sorry, fixed now

mamoutou-diarra (approved):

AlbertoSoutullo (approved):
LGTM thanks!

PR #367, opened 2026-07-25
Add extra values to metadata for time intervals

PearsonWhite:
Add Grafana link, VictoriaLogs link, Duration to metadata subdicts that contain "start" and "end" values.

### DIFF
--- a/src/deployments/core/base_bridge.py
+++ b/src/deployments/core/base_bridge.py
@@ -10,9 +10,9 @@ from pydantic import BaseModel
 from src.deployments.core.event_log import find_events, parse_events_log
 from src.deployments.core.event_mapping import EventMapping
 from src.deployments.core.metadata_times import (
+    apply_time_shifts,
     enrich_intervals,
     format_metadata_timestamps,
-    get_valid_shifted_times,
 )
 
 logger = logging.getLogger(__name__)
@@ -35,8 +35,7 @@ class BaseBridge(BaseModel):
         metadata = parse_events_log(events_log_path, events_maps)
 
         deltatime_map = {obj.target: obj.time_shift for obj in events_list}
-        shifted = get_valid_shifted_times(deltatime_map, metadata)
-        metadata.update(shifted)
+        metadata = apply_time_shifts(deltatime_map, metadata)
 
         metadata = enrich_intervals(metadata, namespace=namespace)
         metadata = format_metadata_timestamps(metadata, format="vquery")
--- a/src/deployments/core/base_bridge.py
+++ b/src/deployments/core/base_bridge.py
@@ -17,7 +17,7 @@ from src.deployments.core.metadata_times import (
     grafana_link,
     victorialogs_link,
 )
-from src.utils.dict_utils import dict_apply
+from src.utils.dict_utils import dict_transform
 
 logger = logging.getLogger(__name__)
 
@@ -60,10 +60,10 @@ class BaseBridge(BaseModel):
 
         if grafana:
             func = partial(apply_args, key="grafana", func=grafana_link)
-            metadata = dict_apply(metadata, func)
+            metadata = dict_transform(metadata, func)
         if vlogs:
             func = partial(apply_args, key="victoria_logs", func=victorialogs_link)
-            metadata = dict_apply(metadata, func)
+            metadata = dict_transform(metadata, func)
 
         def apply_duration(metadata, key):
             try:
@@ -75,7 +75,7 @@ class BaseBridge(BaseModel):
 
         if duration:
             duration_func = partial(apply_duration, key="duration")
-            metadata = dict_apply(metadata, duration_func)
+            metadata = dict_transform(metadata, duration_func)
 
         return metadata
 
--- a/src/deployments/core/base_bridge.py
+++ b/src/deployments/core/base_bridge.py
@@ -1,8 +1,10 @@
 # Python Imports
 import logging
 from collections import defaultdict
+from copy import deepcopy
+from functools import partial
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 
 from pydantic import BaseModel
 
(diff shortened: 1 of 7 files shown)


src/deployments/core/metadata_times.py
```diff
@@ -57,6 +58,69 @@ def collect_valid_interval(path, obj):
     return filtered
 
 
+def grafana_link(start_time, end_time, namespace=None):
+    base = "https://grafana.lab.vac.dev/d/jIrqsZTIz/nwaku"
+    params = {
+        "orgId": 1,
+        "from": start_time,
+        "to": end_time,
```
AlbertoSoutullo: Add typing to parameters, because this can be wrongly used by the way this is implemented. ATM `start_time` and `end_time` are still `datetime` objects, `urlencode()` will stringify them like `2026-01-01 12:00:00` and then `format_timestamp_url()` will add a `+` in the space.
The PR adds a test that only checks the URL prefix and namespace but it doesn't verify the actual from and to parameters. This could produce links that Grafana interprets incorrectly.

src/deployments/core/event_window_bridge.py
```diff
@@ -39,8 +42,8 @@ def get_metadata(self, events_log_path: Path) -> dict:
 
         try:
             selected_interval = events[self.interval]
-            metadata["stack"]["start_time"] = selected_interval["start"]
-            metadata["stack"]["end_time"] = selected_interval["end"]
+            metadata["stack"]["start_time"] = selected_interval.get("start", EventNotFound)
+            metadata["stack"]["end_time"] = selected_interval.get("end", EventNotFound)
```
AlbertoSoutullo: Why are we useing `.get()` with a default value in a try except here? This means we can use `start_time` as `EventNotFound`. Why do we want that? I don't see this is handled elsewhere, and I think it should be handled here.

src/deployments/core/base_bridge.py
```diff
@@ -12,19 +14,79 @@
...
+            func = partial(apply_args, key="victoria_logs", func=victorialogs_link)
+            metadata = dict_apply(metadata, func)
+
+        def apply_duration(metadata, key):
+            try:
+                new_metadata = deepcopy(metadata)
+                new_metadata[key] = format_duration(metadata["end"] - metadata["start"])
+                return new_metadata
+            except Exception:
+                return new_metadata
```
AlbertoSoutullo: I don't understand this. Why are we returning the `new_metadata` if it caused an exception? How do we differenciate if it has valid data or not?

src/utils/dict_utils.py
```diff
@@ -131,7 +131,7 @@ def apply(path, node):
             if path == Path():
                 new_dict = new_value
             else:
-                dict_set(new_dict, path, new_value)
+                dict_set(new_dict, path, new_value, replace_leaf=True)
```
AlbertoSoutullo: I can see why this is needed, but this changes the behavior of `dict_apply()` for every existing caller.
Either we do this elsewhere or this has to be a flag default to false.

AlbertoSoutullo (changes requested):

src/deployments/core/event_window_bridge.py
PearsonWhite: I added this because if the experiment never finished, there won't be an end time, but we still may want to create a metadata.json that includes the start time.

For example, I has some experiments that ran fine, but after publishing something failed, so `internal_run_finished` never appeared in `events.log`. I manually created the Bridge and processed the events, but because there was no end time, the exception (Missing {self.interval}...) was raised instead of generating the metadata. This change allowed me to generate the metadata even with one end of the window missing.

src/deployments/core/base_bridge.py
PearsonWhite: This is because of how `dict_apply` works: by applying `func` to every single element, and then building a new dict using those elements. If `None` is returned, it ignore the value at that path. Returning the deepcopied value will preserve the non-interval elements.

src/utils/dict_utils.py
PearsonWhite: Not replacing leaf nodes by default is unintuitive, so I made it the default in 9fde1e43e93ce20069200e13ac238bbcfb4b3f2d

AlbertoSoutullo: Yeah but default value should remain False as this affects every existing previous caller.

src/deployments/core/tests/test_base_bridge_split.py
```diff
@@ -136,3 +141,50 @@ def test_base_bridge_preserves_full_metadata_flow_with_event_mappings(tmp_path):
...
+
+
+def test_format_metadata_timestamps_vquery_and_url():
+    metadata = {"root": {"start": datetime(2026, 1, 1, 12, 0, 1, 123456)}}
+
+    result_vq = format_metadata_timestamps(metadata, "vquery")
+    result_vq["root"]["start"] == "2026-01-01T12:00:01"
+
+    result_url = format_metadata_timestamps(metadata, "url")
+    result_url["root"]["start"] = "2026-01-01T12:00:01.123Z"
```
AlbertoSoutullo: This test is not asserting anything

src/deployments/core/event_window_bridge.py
AlbertoSoutullo: But `_get_metadata_events()` calls `get_valid_shifted_times()`, which only emits intervals containing valid data. So if an experiment has only a start event and no end event, the interval is removed entirely from events. `events[self.interval]` then raises `KeyError`, so the new `EventNotFound` values are never reached.

src/deployments/core/base_bridge.py
AlbertoSoutullo: I think this whole `apply_extras_func` is confusing and should be refactored:
- It traverses the entire metadata tree up to three times
- `apply_args()` and `apply_duration()` duplicate the same pattern.
- partial() makes the control flow harder to follow.
- except `Exception` silently hides programming errors.
- Each extra is applied independently, so an interval can be partially modified if one calculation fails.

AlbertoSoutullo (changes requested):

src/deployments/core/tests/test_base_bridge_split.py
PearsonWhite: Fixed in 76b4f6bc8e838eb66117e44638199c777eb08ec0

AlbertoSoutullo: Yes, but the test is not asserting

src/deployments/core/event_window_bridge.py
PearsonWhite: `get_valid_shifted_times` returns `shifted` which is then merged into `metadata`, so it need not return anything for incomplete intervals to end up in the metadata. The logic could be improved to avoid `update`, but it's beyond the scope of this PR.

src/deployments/core/base_bridge.py
PearsonWhite: Addressed in 76b4f6bc8e838eb66117e44638199c777eb08ec0

src/utils/dict_utils.py
PearsonWhite: Other changes make it such that we don't need this change, so I reverted it in 647fc3b7cd419d2e5c22384e6e8cd8389bc68a1d

src/deployments/core/tests/test_base_bridge_split.py
PearsonWhite: Fixed in https://github.com/vacp2p/10ksim/commit/76b4f6bc8e838eb66117e44638199c777eb08ec0

src/deployments/core/metadata_times.py
```diff
@@ -68,3 +133,62 @@ def add_links(metadata, links_map):
...
+    Otherwise: leave unchanged.
+    """
+    if not _is_interval_node(node):
+        raise KeepNode()
+
+    start = node["start"]
+    end = node["end"]
+
+    if end < start:
+        raise KeepNode()
```
AlbertoSoutullo, line 155: Shouldn't this be `<=` ?

src/deployments/core/event_window_bridge.py
AlbertoSoutullo: Hm ok. But I is still confusing to me why we want to know the end time of a failed experiment.

AlbertoSoutullo (changes requested):

src/deployments/core/metadata_times.py
PearsonWhite: We can still calculate duration and links even if the start and end times are the same.

src/deployments/core/event_window_bridge.py
PearsonWhite: What this change does is: instead of trying to get `selected_interval["start"]`, failing and raising `ValueError(f"Missing `{self....`, which would prevent generating the metadata at all, we allow it to continue and just make the field. So we would end up with a metadata.json with 
```
{
     "stack": {
          ...
          "start_time": "2026-07-29T14:30:34",
          "end_time": "EventNotFound",
          ...
```
I could revert this if you want. I just think it's easier to work with having the metadata.json so we can inspect the data it was able to create. For example, we can still grab the start time and look in grafana starting there.

AlbertoSoutullo: But this doesn't happen in the current implementation. Unless I am missing something, if the log contains only complete.start:

1. `parse_events_log()` produces something like `{"complete": {"start": ...}}`.
2. `get_valid_shifted_times()` drops `complete` because it has no `end`.
3. `_get_metadata_events()` returns `{}`.
4. `events[self.interval]` raises `KeyError`.
5. The existing `ValueError` is still raised.

PearsonWhite: I separated the logic out so shifting times for events is independent of "valid interval" logic in 631cebe124f8bb6c659924a609ae3ccfeec2d0b7
Also, in c9c14014d2f8dffb6ac19b62bdafece85de355ab I updated the logic so even with a missing interval we create the metadata.

So the logic will be:
1. parse_events_log() produces something like {"complete": {"start": ...}}.
2. metadata = apply_time_shifts(deltatime_map, metadata) # applies time shifts to metadata
3. get_metadata creates interval with EventNotFound for end_time.

AlbertoSoutullo (approved):
I think thats all

PR #369, opened 2026-07-28
Use local rng for parametrized logoscore test

PearsonWhite:
Otherwise it interferes with the global random seed, even when the main program is being run instead of the tests.


AlbertoSoutullo (approved):
Oh yeah I missed this. Seeds should always be used in tests, thats for sure.

PR #371, opened 2026-07-31
Scrape the settled window for Shadow runs

radiken:
Cluster runs scrape the `stable` window, but Shadow logs none of the events it keys on, so it scraped the whole run and the gauges carried mesh formation and teardown. Anchors on the received counter leaving zero instead, with the same offsets.


AlbertoSoutullo (commented):
Are we sure this works?
For the default configuration:

- publisher starts at approximately 90 seconds
- first delivery will usually occur at or after 90 seconds
- start_epoch = first_delivery + 180
- the imported data ends at approximately 180 seconds, minus 30 seconds 

This will end up with an empty inveral, ie:
start = 90 + 180 = 270s
end   = 180 - 30  = 150s

AlbertoSoutullo (commented):
We should note somewhere that this (as far as I understand) will only work properly if we are using one topic. Am I right?

PR #372, opened 2026-08-02
Add Vaclab Nodes Resources and Topology to DST_DASHBOARD

mamoutou-diarra:
## Context
This PR adds vaclab node resources (CPU, RAM, DISK, BW) and network topology to the DST dashboard. The resources are fetched using prometheus api and the topology is a static star network (representing all nodes connected via the same overlay).

## Main Changes
- Added vaclab processors, to fetch node related prometheus metrics and transform them into key/value
- Added vaclab /nodes api endpoint that returns resources and a static network diagram
- Updated Network and Resources page in Frontend to render resource table and network topology

## URL
- Already deployed in DEV:  [https://dev.dashboard.lab.vac.dev](https://dev.dashboard.lab.vac.dev)


dst_dashboard/processors/vaclab_processor.py
```diff
@@ -0,0 +1,264 @@
...
+                    results[key] = []
+                    continue
+
+                match outcome:
+                    case Ok(vector):
+                        results[key] = vector
+                    case Err(reason):
+                        logger.warning(f"Prometheus query '{key}' returned no data: {reason}")
+                        results[key] = []
+        return results
```
AlbertoSoutullo, line 198: This converts every failed query into an empty list, and `get_snapshot()` then returns a successful snapshot with nodes: []. Any VictoriaMetrics failed uname query will therefore make the UI display “0 nodes” instead of reporting that the data source is unavailable.
This is especially problematic because the frontend treats a successful response as valid data and does not show an error when data.nodes is empty.
We need to correctly propagate those errors.

AlbertoSoutullo (changes requested):

dst_dashboard/frontend/src/App.js
```diff
@@ -18,19 +20,15 @@ function App() {
...
-                                    element={<ComingSoonPage title="Networks" description="Live network explorer is coming soon." />}
-                                />
-                            </Routes>
+                            <PageTransition>
+                                <Routes>
+                                    <Route path="/" element={<HomePage />} />
+                                    <Route path="/experiments" element={<ExperimentsPage />} />
+                                    <Route path="/experiment/:experimentId" element={<ExperimentPage />} />
+                                    <Route path="/vaclab/topology" element={<TopologyPage />} />
+                                    <Route path="/vaclab/networks" element={<ResourcesPage />} />
```
radiken, line 29: The page names are a bit confusing to me, topology page is titled "Network" and `/vaclab/networks` renders "Resources", probably worth matching the names

PR #375, opened 2026-08-06
Stop dropping deliveries whose measured delay is negative

radiken:
Sometimes the delivery delay is shown as -1 or -2, because of the clock skew between the servers. 

`delayMs` was matched with `[\w*]+`, which excludes the minus sign, so deliveries recorded with a negative delay were silently discarded. Affects nimlibp2p and gossipsub-priority-queues, cluster and Shadow, since both use `Nimlibp2pTracer`.


PearsonWhite (commented):
Did you check the Nimlibp2pAnalyzer to see if the logic accounts for potentially negative times?

AlbertoSoutullo (commented):
I don't really think this is the correct approach. We can get rid of this just using a small artificial delay. As long as we take that into consideration for report results when talking about latency, we can just do experiments with 30ms or 50ms and this will never happen. Accept negative for analysis numbers should never be the case.

AlbertoSoutullo (approved):

PR #377, opened 2026-08-06
Scrape the settled window for Shadow runs

radiken:
Cluster runs scrape the `stable` window, but Shadow logs none of the events it keys on, so it scraped the whole run and the gauges carried mesh formation and teardown. Anchors on the received counter leaving zero instead, with the same offsets.

Replaces #371 due to an accidental close, the review discussion on #371 still applies.

### DIFF
--- a/src/analysis/metrics/shadow_metrics.py
+++ b/src/analysis/metrics/shadow_metrics.py
@@ -95,6 +95,23 @@ class EphemeralVictoriaMetrics:
         logger.info(f"Removed ephemeral VictoriaMetrics `{self._name}`")
 
 
+STABLE_OFFSET_S = 180  # same offsets as the bridge's `stable` interval
+STABLE_TAIL_S = 30
+RECEIVED_METRIC = "libp2p_gossipsub_received_total"
+
+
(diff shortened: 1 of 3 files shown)


src/analysis/metrics/shadow_metrics.py
```diff
@@ -95,6 +95,34 @@ def __exit__(self, *exc) -> None:
         logger.info(f"Removed ephemeral VictoriaMetrics `{self._name}`")
 
 
+STABLE_OFFSET_S = 180  # same offsets as the bridge's `stable` interval
+STABLE_TAIL_S = 30
```
AlbertoSoutullo: I think these should be imported from the original place rather than re-defined

AlbertoSoutullo (changes requested):

src/analysis/metrics/shadow_metrics.py
radiken: Fixed in [7705f28](https://github.com/vacp2p/10ksim/pull/377/commits/7705f280978ecf8e83e1ba398410aa61446e78c8)

AlbertoSoutullo (approved):

PR #378, opened 2026-08-06
Files for logos-blokchain testnet

AlbertoSoutullo:
To deploy what is mentioned in the release steps: https://github.com/logos-blockchain/logos-blockchain/releases/tag/0.2.1

### DIFF
--- a/deployment-utilities/kubernetes-utilities/nomos/devnet_0.2.1/logos-blockchain.yaml
+++ b/deployment-utilities/kubernetes-utilities/nomos/devnet_0.2.1/logos-blockchain.yaml
@@ -250,18 +250,17 @@ spec:
                     ;;
                 esac
               }
-          
               wait_for_wallet_note() {
                 LABEL="$1"
                 PUBLIC_KEY="$2"
-          
+  
                 BALANCE_FILE="/tmp/${LABEL}-balance.response"
                 ERROR_FILE="/tmp/${LABEL}-balance.error"
                 STATUS_FILE="/tmp/${LABEL}-balance.status"
-          
+  
                 BALANCE_RESPONSE=""
                 NOTE_ID=""
-          
+  
                 for attempt in $(seq 1 120); do
                   curl -sS \
                     -o "$BALANCE_FILE" \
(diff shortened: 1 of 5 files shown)


deployment-utilities/kubernetes-utilities/nomos/devnet_0.2.1/logos-blockchain.yaml
```diff
@@ -0,0 +1,512 @@
...
+        - name: nomos
+          image: soutullostatus/logos-core-blockchain:v0.2.1
+          imagePullPolicy: Always
+          resources:
+            requests:
+              memory: "1024Mi"
+              cpu: "2000m"
+            limits:
+              memory: "4096Mi"
+              cpu: "4000m"
```
mamoutou-diarra, line 36: if we want Guaranteed QOS and cpu pinning, we need to use the same values in requests and limits. for cpu we can use "2", and for memory "8Gi"

deployment-utilities/kubernetes-utilities/nomos/devnet_0.2.1/logos-blockchain.yaml
```diff
@@ -0,0 +1,512 @@
...
+                  exit 1
+                fi
+
+                awk '
+                  /^  metrics: None$/ {
+                    print "  metrics: !Otlp"
+                    print "    endpoint: \"http://otlp.lab.vac.dev:31317/\""
+                    print "    protocol: \"Grpc\""
+                    print "    service_name: \"" ENVIRON["HOSTNAME"] "\""
+                    print "    authorization_header: \"Bearer ift-f7X2djkoYZkfulDwS9aYITSxo8Q8VDeUb1\""
```
mamoutou-diarra: I think we should hide the token here by using a secret and ENV

mamoutou-diarra (changes requested):
LGTM, but I'm concerned about the bearer token that appears in clear text. We can get it from a secret and just print the $ENV

deployment-utilities/kubernetes-utilities/nomos/devnet_0.2.1/logos-blockchain.yaml
AlbertoSoutullo: Oh my god not again...

deployment-utilities/kubernetes-utilities/nomos/devnet_0.2.1/logos-blockchain.yaml
AlbertoSoutullo: Yes but this is not needed for now its an example node

mamoutou-diarra (approved):

PR #379, opened 2026-08-08
Post-run analysis for the three regression scenarios

radiken:
Each scenario writes its own expectation-vs-result table after the run, with the rows the rulebook lists. Follows #366, which added the scenarios themselves.

### DIFF
--- a/src/analysis/post_run/churn.py
+++ b/src/analysis/post_run/churn.py
@@ -1,0 +1,146 @@
+"""Post-run analysis for the node-churn scenario."""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, List, Optional, Sequence, Tuple
+
+import pandas as pd
+
+from src.analysis.post_run.scenario_common import (
+    POD_COLUMN,
+    Row,
+    load_mesh_peers,
+    mesh_peers_row,
+    pct,
+    prepare,
+    write_table,
+)
+from src.deployments.core.event_log import find_events
+
+if TYPE_CHECKING:
+    from src.deployments.experiments.libp2p.nimlibp2p import NimLibp2pExperiment
+
+logger = logging.getLogger(__name__)
+
+
+def longest_run(mask: Sequence[bool]) -> Tuple[Optional[int], Optional[int], int]:
+    """Start, end and length of the longest True stretch."""
+    best = (None, None, 0)
+    start = None
+    for i, flag in enumerate(list(mask) + [False]):
+        if flag and start is None:
+            start = i
+        elif not flag and start is not None:
+            if i - start > best[2]:
+                best = (start, i - 1, i - start)
+            start = None
+    return best
+
+
+def churn_outages(df: pd.DataFrame, churned: Sequence[str]) -> pd.DataFrame:
+    """Per churned node: the outage read off its own miss run, and what it missed after it.
+
+    Pods keep serving past the scale-down and are back before the rollout finishes, so the
+    scale-down window is not a node's outage; the stretch it missed is.
+    """
+    order = df.drop_duplicates("msgId").sort_values("sent")
+    times = order["sent"].to_numpy()
+    present = (
+        df[df[POD_COLUMN].isin(set(churned))]
+        .pivot_table(index=POD_COLUMN, columns="msgId", values="ordinal", aggfunc="size")
+        .reindex(columns=order["msgId"])
+        .notna()
+    )
+
+    rows = []
+    for pod, got in present.iterrows():
+        missed = ~got.to_numpy()
+        start, end, length = longest_run(missed)
+        after = [i for i, flag in enumerate(missed) if flag and end is not None and i > end]
+        rows.append(
+            {
+                POD_COLUMN: pod,
+                "missed": int(missed.sum()),
+                "outage_messages": length,
+                "outage_s": (
+                    (times[end] - times[start]) / pd.Timedelta(seconds=1) if length else 0.0
+                ),
+                "missed_after": len(after),
+                # How long the node keeps dropping messages once its main outage has ended.
+                "recovery_tail_s": (
+                    (times[after[-1]] - times[end]) / pd.Timedelta(seconds=1) if after else 0.0
+                ),
+            }
+        )
+    return pd.DataFrame(rows)
+
+
+def churn_table(
+    df: pd.DataFrame,
+    churned: Sequence[str],
+    num_messages: int,
+    num_nodes: int,
+    mesh: Optional[pd.DataFrame] = None,
+) -> List[Row]:
+    """Below 100%, with every miss inside the node's own outage."""
+    is_churned = df[POD_COLUMN].isin(set(churned))
+    survivors = num_nodes - len(set(churned))
+    per_survivor = df[~is_churned].groupby(POD_COLUMN)["msgId"].nunique()
+    out = churn_outages(df, churned)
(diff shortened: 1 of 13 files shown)


src/analysis/post_run/churn.py
```diff
@@ -0,0 +1,146 @@
...
+from src.deployments.core.event_log import find_events
+
+if TYPE_CHECKING:
+    from src.deployments.experiments.libp2p.nimlibp2p import NimLibp2pExperiment
+
+logger = logging.getLogger(__name__)
+
+
+def longest_run(mask: Sequence[bool]) -> Tuple[Optional[int], Optional[int], int]:
+    """Start, end and length of the longest True stretch."""
```
AlbertoSoutullo: What "True stretch" means?

src/analysis/post_run/churn.py
```diff
@@ -0,0 +1,146 @@
...
+                    (times[end] - times[start]) / pd.Timedelta(seconds=1) if length else 0.0
+                ),
+                "missed_after": len(after),
+                # How long the node keeps dropping messages once its main outage has ended.
+                "recovery_tail_s": (
+                    (times[after[-1]] - times[end]) / pd.Timedelta(seconds=1) if after else 0.0
+                ),
+            }
+        )
+    return pd.DataFrame(rows)
```
AlbertoSoutullo, line 84: Several comments:
1. I think this description is a bit hard to understand.
2. What do we have in this `df`? Information from `sent` messages, `received` ones or both?
3. Why do we drop duplicates only based on `msgId`? For `received` messages I think it shouldn't matter but for sent this also depents to the peer we are sending. Or am I missing something?
4. Because the pivot index only contains pods that appear in df, a churned node that receives no messages at all is absent from present. Consequently:
    - len(out) is smaller than the number of churned nodes.
    - The report can claim recovery for only the nodes that had at least one delivery.
    - If every churned node receives zero messages, out is empty and later calls such as out["missed_after"].max() can fail.

src/analysis/post_run/churn.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def churn_table(
+    df: pd.DataFrame,
+    churned: Sequence[str],
+    num_messages: int,
+    num_nodes: int,
+    mesh: Optional[pd.DataFrame] = None,
+) -> List[Row]:
+    """Below 100%, with every miss inside the node's own outage."""
```
AlbertoSoutullo: Improve a bit what this function does.

src/analysis/post_run/churn.py
```diff
@@ -0,0 +1,146 @@
...
+            "one contiguous gap, nothing missed either side of it",
+            f"missed {out['missed'].mean():.1f} of {num_messages} on average, "
+            f"{out['outage_messages'].mean():.1f} in one stretch",
+        ),
+        (
+            "clean recovery",
+            "no misses once the node is back",
+            f"{ragged} of {len(out)} nodes keep dropping after their outage, "
+            f"worst {out['missed_after'].max()} messages over "
+            f"{out['recovery_tail_s'].max():.0f}s",
```
AlbertoSoutullo, line 126: If received.csv is empty, or if all churned nodes are absent from the delivery data, out has no rows and the reductions return invalid values or raise. Since post-run analysis is expected to produce a useful report, this should be handled explicitly.

AlbertoSoutullo (commented):
The source of truth for messages should come from the publisher. Are we checking somewhere this? ie: if a message was not published anywhere we should know that.

src/analysis/post_run/churn.py
radiken: t's the longest run of consecutive misses, the mask is per-message "did this node miss it". Reworded now, hope it is clearer.

src/analysis/post_run/churn.py
radiken: 1. Updated.
2. It's the delivery frame, one row per message received by one pod, with sent joined from the publisher. Said so in the comment now.
3. Good question, the dedup is fine here because order is only used to get the publish order of unique messages, and every copy of a message carries the same sent, so dropping on msgId loses nothing, it's not used for counting anything. Added a comment for that.
4. You are right, thanks for catching it! Fixed now by reindexing the rows over the full churned list.

src/analysis/post_run/churn.py
radiken: Good catch, thanks! It's handled in `prepare()` now.

src/analysis/post_run/churn.py
radiken: Expanded it to asserts if delivery is below 100% and every miss is inside the churned node's own outage

src/analysis/post_run/churn.py
```diff
@@ -0,0 +1,155 @@
...
+        .pivot_table(index=POD_COLUMN, columns="msgId", values="ordinal", aggfunc="size")
+        .reindex(index=list(churned), columns=order["msgId"])
+        .notna()
+    )
+
+    rows = []
+    for pod, got in present.iterrows():
+        missed = ~got.to_numpy()
+        start, end, length = longest_run(missed)
+        after = [i for i, flag in enumerate(missed) if flag and end is not None and i > end]
```
AlbertoSoutullo, line 68: I think `after` only records misses whose indexes are greater than the end of the longest missed run.
If a node misses messages in two separate regions ie: one message before the main churn outage and several messages during the outage, `missed_after` remains zero. The report can then claim clean recovery even though the node missed messages outside the selected contiguous outage.

AlbertoSoutullo (changes requested):

AlbertoSoutullo (changes requested):

src/analysis/post_run/partition.py
```diff
@@ -0,0 +1,91 @@
...
+    crossed = int(((under["a"] > 0) & (under["b"] > 0)).sum())
+    full = after[after["reach"] >= num_nodes]
+
+    # The messages before the first full-reach one are still in flight across the merge.
+    if full.empty:
+        reconverged = "no message reached the whole network after the heal"
+    else:
+        reconverged = f"{(full.iloc[0]['sent'] - heal).total_seconds():.0f}s"
+
+    contained = int((under["reach"] == under[["a", "b"]].max(axis=1)).sum())
```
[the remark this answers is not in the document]
radiken: Good catch, thanks! fixed in [927e10c](https://github.com/vacp2p/10ksim/pull/379/commits/927e10c6d7b6497beefecc349fcbc6418f757d9d) and added tests

src/analysis/post_run/churn.py
```diff
@@ -0,0 +1,160 @@
...
+    churned: Sequence[str],
+    num_messages: int,
+    num_nodes: int,
+    mesh: Optional[pd.DataFrame] = None,
+) -> List[Row]:
+    """Expectation-vs-result rows: delivery is below 100%, and every miss sits inside the
+    churned node's own outage rather than trailing after it."""
+    is_churned = df[POD_COLUMN].isin(set(churned))
+    survivors = num_nodes - len(set(churned))
+    per_survivor = df[~is_churned].groupby(POD_COLUMN)["msgId"].nunique()
```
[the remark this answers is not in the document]
radiken: Thanks! Fixed in [df94b78](https://github.com/vacp2p/10ksim/pull/379/commits/df94b78c890e91cc87a43a387a04cf84064160b0)

AlbertoSoutullo (approved):

PR #382, opened 2026-08-08
Fail a Shadow run whose simulated clock has stopped

radiken:
A livelocked Shadow run is currently indistinguishable from a slow one until the job timeout. `wait_for_job_complete` now watches simulated time and fails fast when it stops advancing.

### DIFF
--- a/src/deployments/shadow/runtime.py
+++ b/src/deployments/shadow/runtime.py
@@ -12,6 +12,7 @@ from kubernetes.client.rest import ApiException
 
 from src.deployments.core.k8s_kubeconfig import get_config_file
 from src.deployments.shadow.builders import _RUN_MOUNT, build_log_reader_pod
+from src.deployments.shadow.liveness import StallWatch, simulated_time_s
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/deployments/shadow/runtime.py
```diff
@@ -50,6 +65,25 @@ async def wait_for_job_complete(
...
+            return None
+        name = sorted(pods, key=lambda p: p.metadata.creation_timestamp)[-1].metadata.name
+        return simulated_time_s(
+            core.read_namespaced_pod_log(name=name, namespace=namespace, tail_lines=200)
+        )
+    except Exception as e:
+        # A liveness probe must never be the thing that kills a healthy run, so any
+        # read failure reads as "unknown" and the stall clock keeps its last value.
+        logger.debug(f"Could not read shadow log for liveness check: {e}")
+        return None
```
AlbertoSoutullo, line 84: This returns `None` for both “the clock has not started” and “the log could not be read.” However, `StallWatch.check()` treats every `None` value as a startup condition.
Also this catches all exceptions, including temporary Kubernetes API failures. A healthy run with an extended API outage could be terminated as “stuck before the simulation started.”
I think we should track whether the watch has ever observed simulated time, and treat `None` as startup only before the first valid sample. After startup, `None` should leave the previous liveness timestamp unchanged.

AlbertoSoutullo (changes requested):

AlbertoSoutullo (approved):

PR #383, opened 2026-08-09
Report Shadow mesh state under load, not at the end of the run

radiken:
This PR fixes two problems for shadow regressions to get scrape the correct data. Currently two places reported the idle state: 
1. The settled window ended at the last sample of the run rather than at the end of publishing.
2. The summary read the mesh gauges by last sample.


AlbertoSoutullo (approved):
LGTM nice PR

PR #384, opened 2026-08-11
Read a scrape's metric folder as well as a plain CSV path

radiken:
Lets MetricsPlotter consume a Scrapper dump directly instead of needing the files copied into a different layout first.

### DIFF
--- a/src/analysis/data/data_file_handler.py
+++ b/src/analysis/data/data_file_handler.py
@@ -85,6 +85,24 @@ class DataFileHandler(DataHandler):
 
         return Ok(target_df)
 
+    def _resolve_csv_paths(self, path: Path) -> List[Path]:
+        """A DataPath's CSVs: the file itself, or the ones a scrape wrote inside the folder.
+
+        Scrapper dumps `<location>/<metric folder>/<run name>.csv`, so pointing a DataPath at
+        a scrape dump lands on the metric folder rather than a file.
+        """
+        if path.is_file():
+            return [path]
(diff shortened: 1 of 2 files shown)


src/analysis/data/data_file_handler.py
```diff
@@ -68,6 +68,27 @@ def _concat_data_as_mean_from_file(
...
+                # Dotfiles are never scrape output, and one .DS_Store would take the
+                # whole figure down now that the files are discovered rather than named.
+                csv_names = [name for name in file_names if not name.startswith(".")]
+                if not csv_names:
+                    logger.error(f"{path} holds no files to read.")
+                return sorted(path / name for name in csv_names)
+            case Err(error):
+                logger.error(error)
+                return []
+
```
AlbertoSoutullo, line 91: I was thinking why we are not calling `get_files_from_folder_path` with `*.csv` extension, and then I realized Scrapper writes files without `.csv`.
I think the correct approach here is to write csv files with the suffix csv, so change that in scrapper, and then we can call `get_files_from_folder_path` with the required extension here.

AlbertoSoutullo (changes requested):

src/analysis/data/data_file_handler.py
radiken: #392 is merged and we are now calling `get_files_from_folder_path` with required extension

AlbertoSoutullo (approved):

PR #385, opened 2026-08-11
Shape every run with the WAN profile by default

radiken:
Makes 50ms/50Mbit the default on both platforms, so the code matches what the rulebook already requires.


AlbertoSoutullo (commented):
Hmmm this is what we talked about yesterday in the weekly. Lets wait until @mamoutou-diarra applies the NTP server in the lab before applying a constant latency to see if this has any effect. I would put this changes on hold for now.

Issue #158, opened 2025-10-06
Nim-libp2p regression (recurring)


AlbertoSoutullo:
[Roadmap Link](https://roadmap.vac.dev/dst/vac/2025q4-libp2p-evaluation#regression-testing-recurring)

#### Description
Run different scenarios and collect evidence and data of libp2p's behaviour.

Test for known regressions that have occurred in the past and ensure they don't happen again.

PR #386, opened 2026-08-11
Record the image digest the nodes actually pulled

radiken:
Logs an `images_resolved` event so a run's output says which binary ran, not just which tag was asked for.

Related to #158

### DIFF
--- a/src/deployments/core/k8s_rollout.py
+++ b/src/deployments/core/k8s_rollout.py
@@ -225,6 +225,26 @@ def get_pods_for_statefulset(name: str, namespace: str, api_client=None) -> Iter
     )
 
 
+def resolved_images(name: str, namespace: str, api_client=None) -> dict:
+    """Digest each container of a StatefulSet's pods actually pulled, keyed by container.
+
+    A pod spec holds a mutable tag, so it does not say which binary ran. Values are lists
+    because a mid-run rollout can leave pods on different digests.
+    """
+    digests: dict = {}
+    try:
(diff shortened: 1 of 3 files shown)


AlbertoSoutullo (commented):
Minor coment:  `pod.status` itself can be `None` for a newly created pod, so `pod.status.container_statuses` could raise `AttributeError` and fail the run. Can we check that?

src/deployments/core/k8s_rollout.py
```diff
@@ -239,7 +239,7 @@ def resolved_images(name: str, namespace: str, api_client=None) -> dict:
         return digests
 
     for pod in pods:
-        for status in pod.status.container_statuses or []:
+        for status in getattr(pod.status, "container_statuses", None) or []:
```
AlbertoSoutullo, line 242: `getattr(pod.status, "container_statuses", None)` only protects against the `container_statuses` attribute being absent. It does not protect against `pod.status` itself being `None`. We should evaluate `pod.status` before `getattr()` is called.

AlbertoSoutullo (changes requested):

src/deployments/core/k8s_rollout.py
radiken: It actually covers that case. If `pod.status` is None then `getattr(None, "container_statuses", None)` returns `None` without raising an error.

AlbertoSoutullo (approved):

PR #387, opened 2026-08-11
Bound the publish request and fail a run that lost messages

radiken:
Adds a 30s request timeout and makes a run that failed to publish exit non-zero instead of reporting success.


src/deployments/experiments/base_experiment.py
AlbertoSoutullo, line 1: Hm it would be nice to check why are we doing `self._deployed.clear()`. I don't remember if this is because we can do multi experiments. Because if this is the case, we might need to clear `_failures` as well.
But first, please confirm the reason why we do `clear.`

AlbertoSoutullo (approved):

PR #389, opened 2026-08-11
Capture pod logs with the run as a collector fallback

radiken:
Pulls each pod's log during the post-publish dwell so the analysis does not depend on VictoriaLogs having collected the run.

### DIFF
--- a/src/deployments/core/pod_logs.py
+++ b/src/deployments/core/pod_logs.py
@@ -1,0 +1,56 @@
+"""Pull pod logs straight off the cluster, as a fallback for the log collector.
+
+VictoriaLogs has intermittently shipped nothing for a namespace, which silently breaks
+the whole delivery analysis. Capturing the logs with the run makes it independent of the
+collector, at the cost of the dwell time it takes to pull them.
+"""
+
+import logging
+from concurrent.futures import ThreadPoolExecutor
+from pathlib import Path
+from typing import Optional
+
+from kubernetes import client
+from kubernetes.client import ApiException
+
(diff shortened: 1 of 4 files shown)


src/deployments/core/pod_logs.py
```diff
@@ -0,0 +1,56 @@
...
+            log = api.read_namespaced_pod_log(name=name, namespace=namespace)
+        except ApiException as e:
+            logger.warning(f"Could not read logs from pod `{name}`: {e.status}")
+            return False
+        (dest / f"{name}.log").write_text(log)
+        return True
+
+    names = [pod.metadata.name for pod in pods]
+    with ThreadPoolExecutor(max_workers=workers) as pool:
+        captured = sum(pool.map(capture, names))
```
AlbertoSoutullo: `map` re-raises exceptions from workers. The `capture()` helper only catches `ApiException`, so filesystem errors, unexpected Kubernetes-client errors, or other exceptions will escape and cause `_dwell_and_capture()` to fail.
Please make the per-pod operation failure-isolated, or otherwise explicitly decide which errors should invalidate the run if you consider that should happen.

AlbertoSoutullo (changes requested):

src/deployments/core/pod_logs.py
radiken: Fixed and added tests in [7169a81](https://github.com/vacp2p/10ksim/pull/389/commits/7169a81be678336d491aeebab0a841f65e92b774)

AlbertoSoutullo (approved):

PR #390, opened 2026-08-11
Cross-check log-derived delivery against the nodes' received counter

radiken:
Says whether a delivery shortfall is real or an artefact of log collection, instead of that being a manual step.

### DIFF
--- a/src/analysis/post_run/nimlibp2p.py
+++ b/src/analysis/post_run/nimlibp2p.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
 import logging
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from src.analysis.mesh_analysis.analyzers.data_puller import DataPuller
 from src.analysis.mesh_analysis.analyzers.nimlibp2p_analyzer import Nimlibp2pAnalyzer
 from src.analysis.plotting.latency_plotter import plot_dump_latency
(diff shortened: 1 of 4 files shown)


src/analysis/post_run/nimlibp2p.py
```diff
@@ -25,6 +28,44 @@ def _require_bounded_query(stack: dict) -> None:
...
+    somebody missed, and every node that missed something -- so multiplying them counts
+    deliveries that were never absent.
+    """
+    expected = reliability["expected_num_peers"] * reliability["expected_num_messages"]
+    missing = sum(
+        node.get("missing") or 0
+        for entry in reliability.get("missing_messages", [])
+        for node in entry.get("nodes", [])
+    )
+    return expected - missing
```
AlbertoSoutullo: If one message is absent from every node’s collected logs, it will not appear in missing_messages, so nothing is subtracted for it. Either derive the log count from the number of observed unique messages, or explicitly account for globally absent expected messages.

AlbertoSoutullo (changes requested):

src/analysis/post_run/nimlibp2p.py
radiken: Thanks, fixed in [a8a3d08](https://github.com/vacp2p/10ksim/pull/390/commits/a8a3d08b3f79aa42bed2cd362e8bc7068a9ff492)

AlbertoSoutullo (approved):

PR #391, opened 2026-08-11
Scrape metrics and draw the standard plots in post-run analysis

radiken:
The resource and mesh-health figures now come out of the run instead of a per-campaign script.

Includes #384, which it needs to read a scrape dump; review that one first.

Related to #158

### DIFF
--- a/src/analysis/post_run/metrics.py
+++ b/src/analysis/post_run/metrics.py
@@ -1,0 +1,90 @@
+"""Scrape a finished run's metrics and draw the standard plots.
+
+The reliability and latency analysis already runs itself, but the resource and mesh-health
+half of a regression report did not: every campaign grew its own scrape-and-plot script.
+These are the same figures, drawn the same way, off the run's own stable window.
+"""
+
+import logging
+from pathlib import Path
+from typing import Dict, List, NamedTuple, Optional
+
+from src.analysis.data.data_file_handler import DataPath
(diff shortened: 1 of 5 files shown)


src/analysis/post_run/metrics.py
```diff
@@ -0,0 +1,90 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class PlotSpec(NamedTuple):
+    name: str
+    metrics: List[str]
+    ylabel: str
+    scale: int
+    fig_size: List[int]
```
AlbertoSoutullo: Is there any reason why we are using this and not `src/analysis/plotting/config.py`?

Because I have the feeling we are duplicating code here

radiken: You are right. It has been a while since I created this but looking at this now I have no idea why I didn't use `PlotConfig`. Dropped it now in [af1dee5](https://github.com/vacp2p/10ksim/pull/391/commits/af1dee58d506555cd032511e24d1a2db04d03250)

AlbertoSoutullo (changes requested):
`_has_any()` uses `any(...)`, but `MetricsPlotter` processes every metric in `template.metrics` for every selected group. If a dump contains `libp2p-in` but not `libp2p-out`, it is included for the bandwidth plot and then fails when the missing metric produces an empty dataframe.
We shoudl make sure that all metrics exist for the plot, becuase if something is missing, it means that something weird happened.

